PR #27, opened 2019-11-11
Add handlers for when the rewriting is finished.

gabi-250:
This should enable users to add a handler that is called after the rewriter finishes.


inikulin:
@gabi-250 this requires lol-rebase now

gabi-250:
@inikulin done

PR #37, opened 2019-11-26
Test that all the tools compile.

gabi-250:
* This fixes a compile error in `tools/parser_trace`:

```
error[E0599]: no method named `as_bytes` found for type `&[u8]` in the current scope
   --> cool-thing/src/base/debug_trace.rs:48:44
    |
48  |                 let mut chunk_str = $chunk.as_bytes().as_debug_string();
    |                                            ^^^^^^^^
    |
   ::: cool-thing/src/parser/mod.rs:122:21
    |
122 |                     trace!(@continue_from_bookmark sm_bookmark, self.current_directive, input);
    |                     --------------------------------------------------------------------------- in this macro invocation

error: aborting due to previous error
``` 

* `scripts/test.sh` now also tests that the code in `tools` builds successfully

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -9,5 +9,9 @@ cargo test --features=integration_test "$@"
 echo "=== Running C API tests... ==="
 prove -e 'cargo' run ::  --manifest-path=./c-api/tests/Cargo.toml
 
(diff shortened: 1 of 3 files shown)


inikulin:
@gabi-250 Thanks!

Issue #38, opened 2019-12-01
Implement WASM-based JS API 


inikulin:

GuyLewin:
Thanks for releasing this project open-source! Is there any ETA regarding this issue? It's an awesome project and I'd love to explore integration with my JS project.

inikulin:
@GuyLewin Thank you for your interest in the project. There is no ETA yet, but I'll probably start working on it in next two weeks.

benjamingr:
This is awesome 👏. Would love to use it.

benjamingr:
Any way I can help promote or push this?

inikulin:
@benjamingr Thanks for your willingness to help! The main functionality is almost implemented, so the biggest chunk of work left is to write tests for everything and documentation. I would appreciate any help with that.

benjamingr:
That's tricky since I don't actually understand how to use it but once there are docs I would be happy to contribute some tests :)

coreybutler:
Has there been any progress on this? 

I've been building an updated variation of the cloudworker app (see https://github.com/dollarshaveclub/cloudworker/issues/138) and it seems like having a WASM release would make it possible to support the worker functionality locally. It'd be excellent to see a pre-built WASM file as part of the next release.

toinbis:
Hi,

could you maybe provide a simple example of how to integrate lol-html into node.js project with just simple command line output parsing? not the most elegant solution but would allow to incorporate lol-html in the test suits. thanks!

ukd1:
Also need this so I can test workers before deploying 👍

cdloh:
Has this been added for this quarter by chance? As with everyone else it'd be great to get this in place for local testing.

yacinehmito:
Adding my voice to the fray. I'd love to have this. If you need help on tests and documentation, I have a couple of free days next week.

mrbbot:
Hey everyone 👋, I wanted to use this for [Miniflare](https://github.com/mrbbot/miniflare)'s `HTMLRewriter`, and I've now got something that works, albeit with some caveats.

---

The first step was to build on @inikulin's great work implementing a JS API. I made the following changes:

- Changed how attributes were returned for element handlers to match the API here: https://developers.cloudflare.com/workers/runtime-apis/html-rewriter#properties. `attributes` now returns an array of `[name, value]` pairs as opposed to an object with `name`/`value` keys, and `get_attribute` now returns `null` instead of `undefined` if an attribute isn't found.
[...]

PR #41, opened 2019-12-25
Implement JS API, part 1: rewritable units (#38)

inikulin:


sejoker:
do you plan to add tests using https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/browsers.html as a part of this PR? It's hard to know whether API are covers everything we need without any kind of tests.

inikulin:
@sejoker Yes. This is just part one, which can't really be tested in the browser yet as there is no way to manually create rewritable units.

Just wanted to do that in parts, otherwise the PR would be huge. Tests will be added once I implement bindings for the rewriter itself.

js-api/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "lol-html-js-api"
+version = "0.1.0"
+authors = ["Ivan Nikulin <inikulin@cloudflare.com>"]
+edition = "2018"
+
+[dependencies]
+lol_html = { path = "../" }
+serde = { version="1.0.104", features = ["derive"] }
```
xtuc: nit
```suggestion
serde = { version = "1.0.104", features = ["derive"] }
```

inikulin: Thanks!

PR #47, opened 2020-06-09
Add support for :nth-child selectors

### LINKED ISSUE #44: :nth-child
Hi there, 

we're trying to generate css selectors automatically and use them with HTMLRewriter (on Cloudflare workers) / lol-html. 
For a guaranteed selector generation we need to use :nth-child selectors in some cases. I completely understand that this could be impossible due to the streaming nature of lol-html. 
Did you approach (and maybe have a workaround) for such a case?

Thanks and all the best

Alex

ObsidianMinor:
Fixes #44.

This adds support for 4 more selectors
  * :nth-child
  * :first-child
  * :nth-of-type
  * :first-of-type

This does not add support for :last-child or :last-of-type since they aren't easily implementable in a streaming rewriter as a single selector has the ability to cause large amounts of buffering.

This PR also cleans up some other selector VM code to make it a bit more readable (and save some memory).


src/selectors_vm/compiler.rs
```diff
@@ -23,46 +31,134 @@ pub struct CompiledAttributeExprOperand {
...
+
+        #[inline]
+        fn compile_literal_lowercase(encoding: &'static Encoding, lit: &str) -> Result<Bytes<'static>, HasReplacementsError> {
+            compile_literal(encoding, &lit.to_ascii_lowercase())
         }
+
+        attribute_exprs.push(
+            match simple_expr {
+                OnAttributesExpr::Id(id) =>
+                    compile_literal(encoding, id)
```
inikulin, line 118: The pattern:
``` rust
  <compile>
       .map_or_else::<CompiledAttributeExpr, _, _>(
               |_| compile_expr!(@unmatchable),
               ...smthg else)
        ),
```
is repeated multiple times and can be refactored.

src/selectors_vm/compiler.rs
```diff
@@ -23,46 +31,134 @@ pub struct CompiledAttributeExprOperand {
...
+                OnAttributesExpr::AttributeExists(name) =>
+                    compile_literal(encoding, name)
+                        .map_or_else::<CompiledAttributeExpr, _, _>(
+                            |_| compile_expr!(@unmatchable),
+                            |name| compile_expr!(|_, m| m.has_attribute(&name))
+                        ),
+
+                &OnAttributesExpr::AttributeExpr(AttributeExpr { ref name, ref value, case_sensitivity, operator }) => {
+                    #[inline]
+                    fn compile_operands(encoding: &'static Encoding, name: &str, value: &str) -> Result<(Bytes<'static>, Bytes<'static>), HasReplacementsError> {
```
inikulin, line 140: Get rid of nested functions - they make code extremely hard to read. Ideally, as a rule of thumb functions should have such a length that it should be possible to read its code on 13 inch display without scrolling.

src/selectors_vm/compiler.rs
```diff
@@ -23,46 +31,134 @@ pub struct CompiledAttributeExprOperand {
...
+                    *flags |= ProgramFlags::NTH_OF_TYPE;
+                    compile_expr!(|state, _| state.type_counter.expect("Counter for type required at this point").is_nth(nth))
+                }
+            }
+        );
+    }
+}
+
+impl Compilable for Expr<OnAttributesExpr> {
+    fn compile(
```
inikulin, line 96: I find this variant of the function much harder to read. What we want to achieve with this refactoring?

src/selectors_vm/compiler.rs
```diff
@@ -23,46 +31,134 @@ pub struct CompiledAttributeExprOperand {
...
+        flags: &mut ProgramFlags,
+    ) {
+        let &Self { negation, ref simple_expr } = self;
         curry_compile_expr_macro!(negation);
 
-        match self {
-            Ok(v) => compile(v),
-            Err(_) => compile_expr!(@unmatchable),
+        local_name_exprs.push(
+            match simple_expr {
```
inikulin, line 68: Avoid deep nesting of expressions: it significantly hurts readability of the code. Here result of the match arm can be assigned to a variable which later pushed to exprs vector. This allow to get rid of one nesting level.

src/selectors_vm/ast.rs
```diff
@@ -1,86 +1,120 @@
...
-    AttributePrefix(AttributeExprOperand),
-    AttributeSubstring(AttributeExprOperand),
-    AttributeSuffix(AttributeExprOperand),
+impl NthChild {
+    /// A first child with a step of 0 and an offset of 1
+    pub fn first() -> Self {
+        Self::new(0, 1)
+    }
+
+    pub fn new(step: i32, offset: i32) -> Self {
```
inikulin, line 20: `first` and `new` should be `#[inline]`

src/selectors_vm/ast.rs
```diff
@@ -1,86 +1,120 @@
...
+        Self { step, offset }
+    }
+
+    pub fn has_index(self, index: i32) -> bool {
```
inikulin, line 24: It would be nice to have a unit test for this. This one of those arithmetic code parts that are easy to break.

src/selectors_vm/ast.rs
```diff
@@ -1,86 +1,120 @@
...
+            .field("name", &self.name)
+            .field("value", &self.value)
+            .field("case_sensitivity", &self.case_sensitivity)
+            .field("operator", match self.operator {
+                AttrSelectorOperator::Equal     => &"AttrSelectorOperator::Equal",
+                AttrSelectorOperator::Includes  => &"AttrSelectorOperator::Includes",
+                AttrSelectorOperator::DashMatch => &"AttrSelectorOperator::DashMatch",
+                AttrSelectorOperator::Prefix    => &"AttrSelectorOperator::Prefix",
+                AttrSelectorOperator::Substring => &"AttrSelectorOperator::Substring",
+                AttrSelectorOperator::Suffix    => &"AttrSelectorOperator::Suffix",
```
inikulin, line 78: This doesn't seem to be formatter with `cargo fmt`

src/selectors_vm/ast.rs
```diff
@@ -1,86 +1,120 @@
...
+#[derive(Eq, PartialEq)]
+pub struct AttributeExpr {
+    pub name: String,
+    pub value: String,
+    pub case_sensitivity: ParsedCaseSensitivity,
+    pub operator: AttrSelectorOperator,
 }
 
-impl SimpleExpr {
+impl AttributeExpr {
```
inikulin, line 54: This name clashes with `OnAttributeExpr`. Probably we can rename it to something like `AttributeExprBinaryOpExpr`

src/selectors_vm/ast.rs
```diff
@@ -1,86 +1,120 @@
...
 
-        match operator {
-            DashMatch => AttributeExpr::AttributeDashMatch(operand),
-            Equal => AttributeExpr::AttributeEqual(operand),
-            Includes => AttributeExpr::AttributeIncludes(operand),
-            Prefix => AttributeExpr::AttributePrefix(operand),
-            Substring => AttributeExpr::AttributeSubstring(operand),
-            Suffix => AttributeExpr::AttributeSuffix(operand),
-        }
+impl Debug for AttributeExpr {
```
inikulin, line 66: Can't we just use `derive` here?

src/selectors_vm/ast.rs
ObsidianMinor: There is one added at the bottom (though I should add more case tests).

src/selectors_vm/ast.rs
ObsidianMinor: Sadly we can't since `AttrSelectorOperator` doesn't implement `Debug` for some reason.

src/selectors_vm/program.rs
```diff
@@ -18,244 +19,96 @@ where
...
-    ) -> Box<Self> {
-        Box::new(GenericInstr {
-            associated_branch,
-            non_attr_exprs,
-            attr_exprs,
-        })
+bitflags! {
+    pub struct ProgramFlags: u16 {
+        /// Enables nth-of-type tag tracking.
+        const NTH_OF_TYPE = 0b0000_0001;
```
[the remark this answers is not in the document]
ObsidianMinor: Would an enum also work?

src/selectors_vm/stack.rs
```diff
@@ -33,11 +35,28 @@ pub enum StackDirective {
...
+pub struct ChildCounter {
+    cumulative: i32,
+}
+
+impl ChildCounter {
+    pub fn add_child(&mut self) {
+        self.cumulative += 1;
+    }
+
+    pub fn is_nth(&self, nth: NthChild) -> bool {
```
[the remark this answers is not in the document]
ObsidianMinor: Is there a general guideline I should follow when applying inline to functions? Most functions like this get inlined automatically.

src/selectors_vm/compiler.rs
```diff
@@ -82,187 +178,95 @@ where
...
-            }
+        Predicate {
+            on_tag_name_exprs,
+            on_attr_exprs,
+        }: &Predicate,
+        branch: ExecutionBranch<P>,
+        flags: &mut ProgramFlags,
+    ) -> Instruction<P> {
+        #[inline]
+        fn compile_all<'a, T: Compilable + 'a, I: IntoIterator<Item = &'a T> + 'a>(iter: I, encoding: &'static Encoding, exprs: &mut ExprSet, flags: &mut ProgramFlags) {
```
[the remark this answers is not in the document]
ObsidianMinor: It made more sense to have a separate function when there were more than two expression lists, but now it just makes more sense to repeat the calls.

src/selectors_vm/compiler.rs
ObsidianMinor: It made more sense when there were more ordered expression lists and separate "ops" but now it doesn't seem that worth it now that everything is more or less the same as before.

ObsidianMinor: Actually, I just noticed how we can remove the "curry" and "compile" macros which made compilation look a lot more complicated like I originally wanted to do with this refactor.

src/selectors_vm/ast.rs
ObsidianMinor: `cargo fmt` can't format it so I added the extra space to line up everything.

src/selectors_vm/ast.rs
ObsidianMinor: Or `AttributeComparisonExpr` maybe?

src/selectors_vm/program.rs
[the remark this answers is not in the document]
inikulin: If you've had more than one flag - yes. But currently it's just a single toggle.

src/selectors_vm/stack.rs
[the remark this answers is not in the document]
inikulin: In the majority of cases compiler automatically infers that it can inline a function, but now always. E.g. without `#[inline]` compiler won't do cross-crate inlining if LTO is disabled.
Rust stdlib codebase is a good example - `#[inline]` specified explicitly everywhere.

Regarding when to inline: for me the rule of thumb is to inline one-liners and functions with 1-3 expressions in them. Basically, everything what you would write inline by hand in your function, but it's already refactored/extracted somewhere else.

src/selectors_vm/stack.rs
```diff
@@ -62,21 +85,66 @@ impl<'i, E: ElementData> StackItem<'i, E> {
...
         }
     }
 }
 
-pub struct Stack<E: ElementData>(LimitedVec<StackItem<'static, E>>);
+pub struct Stack<E: ElementData> {
+    /// A counter for root elements
+    root_counter: ChildCounter,
+    /// A typed counter for root elements. This is optional to indicate if types are actually being counted.
+    root_typed_counters: Option<HashMap<LocalName<'static>, ChildCounter>>,
```
inikulin, line 100: `root_child_counter`, `root_typed_counters`

inikulin, line 100: `root_child_counter`, `root_typed_counters`

toinbis:
Thanks a lot! Is this going into the cloudflare workers runtime immediately or any time soon?

ObsidianMinor:
Probably next week (or so).

PR #49, opened 2020-06-15
Use more memory efficient nth-of-type tracking

ObsidianMinor:
Previously, a hash map would be allocated for each frame and every local name *that was unique to the frame* would be owned by the map. This new version only requires one hash map/owned unique tag name for all frames on the stack.

This also saves some time when popping the stack back to a previous
local name by using iterators to find the index.

### DIFF
--- a/src/selectors_vm/stack.rs
+++ b/src/selectors_vm/stack.rs
@@ -3,7 +3,8 @@ use super::program::AddressRange;
 use super::ast::NthChild;
 use crate::html::{LocalName, Namespace, Tag};
 use crate::memory::{LimitedVec, MemoryLimitExceededError, SharedMemoryLimiter};
-use hashbrown::{HashSet, HashMap};
+// use hashbrown for raw entry, switch back to std once it stablizes there
+use hashbrown::{HashSet, HashMap, hash_map::RawEntryMut};
 use std::fmt::Debug;
 use std::hash::{Hash, Hasher, BuildHasher};
 
@@ -42,6 +43,11 @@ pub struct ChildCounter {
 
 impl ChildCounter {
     #[inline]
+    pub fn new_and_inc() -> Self {
+        Self { cumulative: 1 }
+    }
+
+    #[inline]
     pub fn inc(&mut self) {
         self.cumulative += 1;
     }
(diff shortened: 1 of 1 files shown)


src/selectors_vm/stack.rs
```diff
@@ -3,7 +3,7 @@ use super::program::AddressRange;
 use super::ast::NthChild;
 use crate::html::{LocalName, Namespace, Tag};
 use crate::memory::{LimitedVec, MemoryLimitExceededError, SharedMemoryLimiter};
-use hashbrown::{HashSet, HashMap};
+use hashbrown::{HashSet, HashMap, hash_map::RawEntryMut};
```
inikulin, line 6: Rust uses hashbrown internally since 1.36, so we can use std HashMap: https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries-8

src/selectors_vm/stack.rs
```diff
@@ -52,13 +52,96 @@ impl ChildCounter {
...
+}
+
+#[derive(Default)]
+/// A more efficient counter that only requires one owned local name to track counters across multiple stack frames
+pub struct TypedChildCounterMap(HashMap<LocalName<'static>, CounterList>);
+
+impl TypedChildCounterMap {
+    /// Adds a seen child to the map. The index is the level of the item
+    pub fn add_child(&mut self, name: &LocalName, index: usize) {
+        let hash = {
```
inikulin, line 75: Nit: move to a separate function instead of closure

src/selectors_vm/stack.rs
```diff
@@ -52,13 +52,96 @@ impl ChildCounter {
...
+        };
+        let entry = self.0.raw_entry_mut().from_hash(hash, |n| name == n);
+        match entry {
+            RawEntryMut::Vacant(vacant) => {
+                let mut counter = ChildCounter::default();
+                counter.inc();
+                vacant.insert_hashed_nocheck(
+                    hash,
+                    name.clone().into_owned(), // the hash won't change just because we've got ownership
+                    CounterList {
```
inikulin, line 88: the majority of the code in this branch can be refactored into `CounterList::new()`

src/selectors_vm/stack.rs
```diff
@@ -52,13 +52,96 @@ impl ChildCounter {
...
+    }
+
+    #[inline]
+    pub fn get<'a, 'i>(&'a self, name: &LocalName<'i>, current: usize) -> Option<&'i ChildCounter>
+    where
+        'a: 'i
+    {
+        match self.0.get(name) {
+            Some(CounterList {
+                current: CounterItem { counter, index },
```
inikulin, line 131: Nit: this clashes with a function argument and a bit confusing. Let's rename the argument

src/selectors_vm/stack.rs
```diff
@@ -159,16 +236,18 @@ impl<E: ElementData> Stack<E> {
...
     ) {
-        for i in (0..self.items.len()).rev() {
-            if self.items[i].local_name == local_name {
-                for item in self.items.drain(i..self.items.len()) {
-                    popped_element_data_handler(item.element_data);
-                }
-
-                break;
-            }
+        let result =
```
inikulin, line 241: Nit: let's rename this variable to better reflect its meaning,  e.g. `last_popped_item_idx` or something like that.

src/selectors_vm/stack.rs
ObsidianMinor: Raw entry APIs are unstable, so we have to explicitly use hashbrown if we want to use its raw entry API.

src/selectors_vm/stack.rs
ObsidianMinor: It's an expression (a closure requires `||`), but sure.

src/selectors_vm/stack.rs
inikulin: Let's leave a comment, so we'll switch to std once it stabilise

inikulin: Though... I might be missing something, but why we can't just use [HashMap::entry()](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.entry) here?

ObsidianMinor: The lifetimes don't check out, so we have to make an owned version of the local name to access the entry even if we don't actually need an owned key to insert.

inikulin: I see, let's just add the comment then.

PR #52, opened 2020-07-01
Use quotes to prevent word splitting.

jbampton:
Checked with ShellCheck.
https://www.shellcheck.net/

https://github.com/koalaman/shellcheck/wiki/SC2046


kornelski:
What if more than one file needs to be added? git porcelain is a multi-line output.

jyn514:
Closing this since it changes the behavior and doesn't seem like of much of a benefit even if it worked.

Issue #56, opened 2020-08-02
Take any `Into<Settings>` for rewrite_str()


jyn514:
Currently, `rewrite_str` only takes a `RewriteStrSettings`, so you can't e.g. configure the memory limit like you could with the full `settings`. It's possible to use `HtmlRewriter` directly, but this is a lot more code and duplicates almost everything in `rewrite_str`. It would be better to take `settings: impl Into<Settings>` in rewrite_str(), which makes the API a lot more flexible.

jyn514:
I'm interested in implementing this.

inikulin:
Any PRs are appreciated.

PR #57, opened 2020-08-02
feat: Allow using either Settings or RewriteStrSettings

jyn514:
Closes https://github.com/cloudflare/lol-html/issues/56


inikulin:
@ObsidianMinor seems like the newest clippy is unhappy with the selector-related code

ObsidianMinor:
See #60 @inikulin

PR #64, opened 2020-09-03
Update to hashbrown 0.9

mbrubeck:
The behavior of the `drain_filter` method has changed in this release:

https://github.com/rust-lang/hashbrown/issues/186

### DIFF
--- a/src/selectors_vm/stack.rs
+++ b/src/selectors_vm/stack.rs
@@ -118,17 +118,17 @@ impl TypedChildCounterMap {
 
     #[inline]
     pub fn pop_to(&mut self, index: usize) {
-        self.0.drain_filter(move |_, v| {
+        self.0.retain(|_, v| {
             while v.current.index > index {
                 match v.items.pop() {
                     Some(next) => {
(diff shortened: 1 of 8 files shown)


src/selectors_vm/stack.rs
```diff
@@ -124,10 +124,10 @@ impl TypedChildCounterMap {
                     Some(next) => {
                         v.current = next;
                     },
-                    None => return false
+                    None => return true
                 }
             }
-            true
+            false
         }).for_each(drop);
```
jyn514, line 131: This could be `.retain()` instead, and then you could switch from an external `hashbrown` dependency to the standard library.

mbrubeck: Switching to `retain` is a good idea for simplicity.  However, this code still can't use the standard library HashMap on stable Rust, because it also uses uses `raw_entry_mut`.

inikulin:
@mbrubeck thanks! sorry for the delay in review

inikulin (approved):

Issue #59, opened 2020-08-02
HtmlRewriter::end() should take `self`, not `&mut self`


jyn514:
[As documented](https://docs.rs/lol_html/0.2.0/lol_html/struct.HtmlRewriter.html#method.write), both `end()` and `write()` will panic if you do anything with the rewriter after you call `end()`. This can be enforced statically by having `end()` consume the writer.

inikulin:
This is how it was implemented initially, but IIRC there were some problems with C API because of that. Need to check.

Issue #46, opened 2020-05-26
Refactor HTMLRewriter Settings to deprecate "try_new"


ObsidianMinor:
`HTMLRewriter`'s creation interface is a bit clunky, since it requires calling `try_new` even if you know that an encoding is valid (for example with `Default`). Encoding should be checked ahead of time via a wrapper type and `try_new` should be deprecated and replaced with an infallible `new` to simplify the API.

See [the API guidelines on static parameter enforcement](https://rust-lang.github.io/api-guidelines/dependability.html#functions-validate-their-arguments-c-validate).

PR #68, opened 2020-10-18
Take `self` in HtmlRewriter::end

jyn514:
>  IIRC there were some problems with C API because of that. Need to check.

This was the error:
```
error[E0507]: cannot move out of `*rewriter` which is behind a mutable reference
  --> src/rewriter.rs:80:31
   |
80 |     unwrap_or_ret_err_code! { rewriter.end() };
   |                               ^^^^^^^^ move occurs because `*rewriter` has type `lol_html::HtmlRewriter<'_, rewriter::ExternOutputSink>`, which does not implement the `Copy` trait
```

This commit combines `end()` and `free()` into the same function for the
C API. It also removes tests that `end()` panics when call twice, and
that `write()` panics when called after `read()`, because they no longer
compile:

```
error[E0382]: borrow of moved value: `rewriter`
   --> src/rewriter/mod.rs:610:13
    |
607 |             let mut rewriter = create_rewriter(512, |_: &[u8]| {});
    |                 ------------ move occurs because `rewriter` has type `rewriter::HtmlRewriter<'_, [closure@src/rewriter/mod.rs:607:53: 607:66]>`, which does not implement the `Copy` trait
608 |
609 |             rewriter.end().unwrap();
    |                      ----- `rewriter` moved due to this method call
610 |             rewriter.write(b"foo").unwrap();
    |             ^^^^^^^^ value borrowed here after move

error[E0382]: use of moved value: `rewriter`
   --> src/rewriter/mod.rs:619:13
    |
616 |             let mut rewriter = create_rewriter(512, |_: &[u8]| {});
    |                 ------------ move occurs because `rewriter` has type `rewriter::HtmlRewriter<'_, [closure@src/rewriter/mod.rs:616:53: 616:66]>`, which does not implement the `Copy` trait
617 |
618 |             rewriter.end().unwrap();
    |                      ----- `rewriter` moved due to this method call
619 |             rewriter.end().unwrap();
    |             ^^^^^^^^ value used here after move
```

As a small performance benefit, `finished` can be removed, making the
rewriter take up slightly less memory.

Closes https://github.com/cloudflare/lol-html/issues/59. I didn't run the Rust tests because of https://github.com/rustation/pre-commit/issues/2, but the C tests all pass.

This is a breaking change.

### DIFF
--- a/c-api/src/rewriter.rs
+++ b/c-api/src/rewriter.rs
@@ -10,6 +10,10 @@ pub struct ExternOutputSink {
     user_data: *mut c_void,
 }
 
+/// This is a wrapper around `lol_html::HtmlRewriter` which allows
+/// use after the rewriter itself is dropped.
+pub struct HtmlRewriter(Option<lol_html::HtmlRewriter<'static, ExternOutputSink>>);
+
 impl ExternOutputSink {
     #[inline]
     fn new(
@@ -39,7 +43,7 @@ pub extern "C" fn lol_html_rewriter_build(
     output_sink: unsafe extern "C" fn(*const c_char, size_t, *mut c_void),
     output_sink_user_data: *mut c_void,
     strict: bool,
-) -> *mut HtmlRewriter<'static, ExternOutputSink> {
+) -> *mut HtmlRewriter {
     use std::convert::TryInto;
 
     let builder = to_ref!(builder);
@@ -57,19 +61,22 @@ pub extern "C" fn lol_html_rewriter_build(
     };
 
     let output_sink = ExternOutputSink::new(output_sink, output_sink_user_data);
-    let rewriter = HtmlRewriter::new(settings, output_sink);
+    let rewriter = lol_html::HtmlRewriter::new(settings, output_sink);
 
-    to_ptr_mut(rewriter)
+    to_ptr_mut(HtmlRewriter(Some(rewriter)))
 }
 
 #[no_mangle]
 pub extern "C" fn lol_html_rewriter_write(
-    rewriter: *mut HtmlRewriter<'static, ExternOutputSink>,
+    rewriter: *mut HtmlRewriter,
     chunk: *const c_char,
     chunk_len: size_t,
 ) -> c_int {
     let chunk = to_bytes!(chunk, chunk_len);
-    let rewriter = to_ref_mut!(rewriter);
+    let rewriter = to_ref_mut!(rewriter)
+        .0
+        .as_mut()
+        .expect("cannot call `lol_html_rewriter_write` after calling `end()`");
 
     unwrap_or_ret_err_code! { rewriter.write(chunk) };
 
@@ -77,10 +84,11 @@ pub extern "C" fn lol_html_rewriter_write(
 }
 
 #[no_mangle]
-pub extern "C" fn lol_html_rewriter_end(
-    rewriter: *mut HtmlRewriter<'static, ExternOutputSink>,
-) -> c_int {
-    let rewriter = to_ref_mut!(rewriter);
+pub extern "C" fn lol_html_rewriter_end(rewriter: *mut HtmlRewriter) -> c_int {
+    let rewriter = to_ref_mut!(rewriter)
+        .0
+        .take() // Using `take()` allows calling `free()` afterwards (it will be a no-op).
+        .expect("cannot call `lol_html_rewriter_end` after calling `end()`");
 
     unwrap_or_ret_err_code! { rewriter.end() };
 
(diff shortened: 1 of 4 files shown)


jyn514:
> This commit combines end() and free() into the same function for the C API.

Note that `free()` still causes memory unsafety if called twice on the same pointer, so if there was anywhere calling `end()` twice before, it will now have UB instead of causing a panic. All the call sites should be audited to make sure they call free() exactly once. This is only relevant for C, since in Rust calling end() twice will be a compile error.

jyn514:
Ooh, I can make this `must_use` too, that way the compiler warns you if you forget to call `end()`.

jyn514:
> Ooh, I can make this must_use too, that way the compiler warns you if you forget to call end().

Hmm, this didn't actually work because `.write()` counts as a 'use'. I wish there was something like `must_consume`.

jyn514:
The closest thing I found was an abandoned discussion on internals.rlo https://internals.rust-lang.org/t/idea-attributes-to-warn-if-value-is-dropped-without-being-acted-upon/10485/19 which suggests a 'proof of work' to make sure `end()` is actually called. That seems like a pretty big burden to force to the user though, really I just want a lint.

jyn514:
A better idea @cuviper thought of is to have `write(self) -> Self` instead of taking `&mut self`; that way whenever you have a Rewriter it will be a fresh expression, so you'll always get the `must_use` lint. That's a pretty invasive change though and will require lots more `ptr::read()`, so probably not worth it, but still worth considering.

inikulin:
If I remember correctly we've tried this approach (with `end()` performing implicit freeing) initially and there were some issue with C++ RAII or something (don't remember all the details tbh). And folks from the Workers team asked me to introduce explicit `free()`. 

The problem with `end()` performing freeing is that it doesn't account for the cases when handlers error and the rewriter becomes poisoned. So, you need to free the rewriter without `end()` ever being called. And in general, I would like to avoid introducing breaking changes in the C API part to not break stuff for Cloudflare Workers.

What we can do instead is consume `self` on Rust side. But for C API use `HtmlRewriterWrapper(Option<HtmlRewriter>)` which derefs to rewriter. We'll keep `free()` which will free the rewriter if it's not `None` at this point (`end()` hasn't be called) or will be noop otherwise. And for `end()` we'll do `self.0.take()`.

jyn514:
> What we can do instead is consume self on Rust side. But for C API use HtmlRewriterWrapper(Option<HtmlRewriter>) which derefs to rewriter. We'll keep free() which will free the rewriter if it's not None at this point (end() hasn't be called) or will be noop otherwise. And for end() we'll do self.0.take().

I like this! Done :)

c-api/src/rewriter.rs
```diff
@@ -52,37 +56,45 @@ pub extern "C" fn lol_html_rewriter_build(
...
 #[no_mangle]
-pub extern "C" fn lol_html_rewriter_end(
-    rewriter: *mut HtmlRewriter<'static, ExternOutputSink>,
-) -> c_int {
-    let rewriter = to_ref_mut!(rewriter);
+pub extern "C" fn lol_html_rewriter_end(rewriter: *mut HtmlRewriter) -> c_int {
+    let rewriter = to_ref_mut!(rewriter)
+        .0
+        .take() // Using `take()` allows calling `free()` afterwards (it will be a no-op).
+        .expect("cannot call `lol_html_rewriter_end` after calling `end()`");
```
inikulin, line 86: I guess it should be "after calling `lol_html_rewriter_free` or `lol_html_rewriter_end`"?

inikulin (approved):
lgtm, with just a minor nit. Thanks!

c-api/src/rewriter.rs
jyn514: If this is called after free(), then one of two things happened:

1. `rewriter.0` was `None` when dropped, so `end()` was called in the past, so the error message is correct (free() is irrelevant to the panic).
2. `rewriter.0` was `Some` when dropped, so this value has already been dropped, so this is UB. In practice, I think the bytes will not be zeroed, so this panic won't execute.

In either case, I don't think there's a need to change the message.

inikulin:
@jyn514 it needs rebase now

jyn514:
Should be ready to go :)

inikulin:
@jyn514 seems like CI's unhappy for some reason

jyn514:
Hmm, that seems related to #46, not this PR ... I fixed it anyway, though.

inikulin:
@jyn514 thanks a lot for all the contributions

PR #70, opened 2020-10-18
Refactor HTMLRewriter Settings to make `HTMLRewriter::new` infallible

jyn514:
- Add `AsciiCompatibleEncoding` wrapper type
- Provide ways to convert between `Ascii...` and `Encoding`
- Make `new` infallible
- Fix tests, examples, and benches

This turned out very well - you only have to worry about fallible
conversions if you use an encoding other than the default, which is
fairly rare.

This is a breaking change.

Closes https://github.com/cloudflare/lol-html/issues/46.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -48,10 +48,9 @@ macro_rules! create_runner {
     ($settings:expr) => {
         move |b, i: &Vec<Vec<u8>>| {
             b.iter(|| {
-                let mut rewriter = lol_html::HtmlRewriter::try_new($settings, |c: &[u8]| {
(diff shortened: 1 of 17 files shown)


jyn514:
This is currently broken because `unwrap_or_ret_err` (unstandably) requires a reasonable error to store in `LAST_ERROR` and right now I'm only giving it `()`. This is fixable by adding a new error type but I'm probably not going to get around to it any time soon.

jyn514:
> This is fixable by adding a new error type

Done, I realized I could reuse the existing `EncodingError`.

inikulin (approved):

Issue #72, opened 2020-10-30
Update `selectors` and `cssparser`


jyn514:
Currently, adding lol-html as a dependency to cargo-deadlinks adds 50 more dependencies (190 -> 240). `cargo tree -d` shows a large number of duplicate dependencies coming from old versions of selectors and cssparser:

```
├── cssparser v0.25.9
│   ├── lol_html v0.2.0 (/home/joshua/src/rust/lol-html)
│   └── selectors v0.21.0
│       └── lol_html v0.2.0 (/home/joshua/src/rust/lol-html)
└── selectors v0.21.0 (*)

phf v0.8.0
└── markup5ever v0.10.0
    └── html5ever v0.25.1
        [dev-dependencies]
        └── lol_html v0.2.0 (/home/joshua/src/rust/lol-html)

phf_codegen v0.7.24
└── cssparser-macros v0.3.6
    └── cssparser v0.25.9 (*)
[build-dependencies]
└── selectors v0.21.0 (*)

phf_codegen v0.8.0
[build-dependencies]
└── markup5ever v0.10.0 (*)

phf_generator v0.7.24
└── phf_codegen v0.7.24 (*)

phf_generator v0.8.0
├── phf_codegen v0.8.0 (*)
└── string_cache_codegen v0.5.1
    [build-dependencies]
    └── markup5ever v0.10.0 (*)

phf_shared v0.7.24
├── phf v0.7.24 (*)
├── phf_codegen v0.7.24 (*)
└── phf_generator v0.7.24 (*)

phf_shared v0.8.0
├── phf v0.8.0 (*)
├── phf_codegen v0.8.0 (*)
├── phf_generator v0.8.0 (*)
├── string_cache v0.8.0
│   └── markup5ever v0.10.0 (*)
└── string_cache_codegen v0.5.1 (*)

rand v0.5.6
[dev-dependencies]
└── lol_html v0.2.0 (/home/joshua/src/rust/lol-html)

rand v0.6.5
└── phf_generator v0.7.24 (*)

rand v0.7.3
├── phf_generator v0.8.0 (*)
└── tempfile v3.1.0
    └── cargo-fuzz v0.8.0
        [dev-dependencies]
        └── lol_html v0.2.0 (/home/joshua/src/rust/lol-html)

rand_chacha v0.1.1
└── rand v0.6.5 (*)

rand_chacha v0.2.2
└── rand v0.7.3 (*)

rand_core v0.3.1
[...]

PR #73, opened 2020-10-30
Update dependencies

jyn514:
Found while looking at https://github.com/cloudflare/lol-html/issues/72. Most commits show how to replicate the change, since they mostly affect the lockfile. This doesn't actually help with #72 since these are all dev-dependencies.


jyn514:
Ugh, this is painful because html5ever removed rcdom in 0.25. I'll just update it to 0.24 for now.

inikulin (approved):

PR #102, opened 2021-07-27
Add esi:include as void element

xtuc:


src/selectors_vm/stack.rs
```diff
@@ -1,27 +1,32 @@
...
         [
             Area, Base, Basefont, Bgsound, Br, Col, Embed, Hr, Img, Input, Keygen, Link, Meta,
             Param, Source, Track, Wbr
         ]
-    )
+    ) {
+        return true;
+    }
+
+    byte_tag_is_one_of!(local_name, encoding, [EsiInclude, EsiComment])
```
inikulin, line 29: You don't need to pass around encoding and decode the tag name: it's all ASCII and all the encodings are ASCII-safe. Just compare local name's bytes, e.g.:
```
if let LocalName::Bytes(bytes) = local_name {
    if bytes == b"esi:include" || bytes == b"esi:comment" {
         return true;
    }
}

```

inikulin (changes requested):

inikulin:
@xtuc for tests take a look at https://github.com/cloudflare/lol-html/blob/master/src/rewriter/mod.rs#L310-L330 You can add similar test there that ensure that `esi:include` is replaced correctly.

src/base/bytes.rs
```diff
@@ -155,6 +155,13 @@ impl<'b> From<&'b [u8]> for Bytes<'b> {
     }
 }
 
+impl PartialEq<[u8]> for Bytes<'_> {
+    #[inline]
+    fn eq(&self, other: &[u8]) -> bool {
+        self.0 == other
+    }
+}
```
inikulin, line 163: `Bytes` deref to `[u8]`, so you don't need that

src/html/tag.rs
```diff
@@ -15,6 +15,15 @@ macro_rules! declare_tags {
...
 
+macro_rules! declare_byte_tags {
+    ($($name:ident = $val:expr),+) => {
+        $(
+            #[allow(non_upper_case_globals)]
+            pub static $name:  &'static [u8] = $val;
+        )+
+    };
+}
+
```
inikulin, line 26: Don't need this macro

src/selectors_vm/stack.rs
```diff
@@ -15,13 +15,23 @@ fn is_void_element(local_name: &LocalName) -> bool {
...
             Area, Base, Basefont, Bgsound, Br, Col, Embed, Hr, Img, Input, Keygen, Link, Meta,
             Param, Source, Track, Wbr
         ]
-    )
+    ) {
+        return true;
+    }
+
+    if let LocalName::Bytes(bytes) = local_name {
+        if bytes == EsiInclude || bytes == EsiComment {
```
inikulin, line 29: Just use byte literals here

inikulin (changes requested):

kornelski (approved):

inikulin (approved):

PR #103, opened 2021-07-28
Update dependencies

kornelski:


src/selectors_vm/parser.rs
```diff
@@ -49,6 +50,14 @@ impl NonTSPseudoClass for NonTSPseudoClassStub {
...
         match *self {}
     }
+
+    fn is_user_action_state(&self) -> bool {
+        match *self {}
+    }
+
+    fn has_zero_specificity(&self) -> bool {
+        match *self {}
+    }
```
xtuc, line 60: I known this isn't new code  but how does it return a bool?

xtuc (approved):

src/selectors_vm/parser.rs
jyn514: This is an enum without any variants, it's impossible for it to be constructed. So if this function is executed, UB has already occurred, so the return type can be anything.

(If it helps, this type could be an alias for std:: Infallible, I'm not sure why it's not.)

PR #114, opened 2021-12-17
Remove Cargo.lock

jyn514:
The Cargo documentation recommends not checking in a Cargo.lock for libraries:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
Additionally, having the lockfile makes it more likely downstream crates will have duplicate dependencies
or incompatible requirements.

This also avoids annoying situations where the pre-commit hook will generate different versions of cargo.lock than the version checked in (e.g. if you have a different default toolchain than the last person who made a commit).


orium (approved):

orium:
There's only two `Cargo.lock` for binaries: tools/parser_trace and tools/selector_ast.  Both are internal tools that are not published, so I don't think it matters much.

Issue #82, opened 2021-03-16
Rewriter reuse example?


jbr:
I'm trying to understand how and if users are intended to express a set of element handlers to be reused on multiple inputs, across multiple threads. In particular, I'm interested in applying a given transformation to a bunch of html byte streams, but the lifetimes and borrows in the public interface make it difficult to see how to achieve that. Is there example code out there, or a part of the documentation that's relevant to this usage pattern?

Thanks!

jyn514:
Hmm, this seems to work fine as-is - I ran `examples/mixed_content_rewriter` on this webpage twice with the same rewriter and it produced two copies of the output. Is there something in particular giving you trouble?

jbr:
The issue is when a user might want to define the rewriting rules on one thread and use them on another (or several others) — because even the rule definitions are !Send, it's unclear how to use this crate in a multithreaded context, or even a `'static` context in which the rules need to be owned somewhere other than where the end user defines them.

[An example](https://docs.rs/trillium-html-rewriter/latest/trillium_html_rewriter/). The workaround I ended up settling on was to have the user provide a `Fn() -> Settings<'static, 'static>` which is run on each thread. It's an awkward interface, but I imagine the end result of this issue is that lol-html is not the appropriate library to use for an async multithreaded html rewriter

jyn514:
It's true that `HtmlRewriter` is `!Send` and `!Sync`, yes. I'm looking into whether it's possible to remove the `Rc<Refcell>` around the output sink.

jyn514:
Oh hmm, I missed that `Settings` is also not Send or Sync, that seems especially unfortunate. I think it would be a breaking change to fix that unfortunately ... the issue is that `ElementContentHandlers` takes a user-provided function and we don't require that to be Send+Sync.

jyn514:
@jbr hmm, I'm confused though - how would it be useful to share the rewriter between threads? How would you synchronize writing the input so you know the output makes sense?

jbr:
There are a few distinct concerns here, and I'm not entirely sure that any of them are specific suggestions or feature requests

## for any multithreaded application

[...]

PR #117, opened 2021-12-21
Make `Settings` Send + Sync

jyn514:
This works by requiring ElementContentHandlers to be Send + Sync. Making `Settings` Send + Sync can be useful if you want to reuse the same settings across multiple rewriters in different threads.

This is a breaking change.

Helps with https://github.com/cloudflare/lol-html/issues/82, but I'd prefer to avoid merging it until I hear back from the author about why they need this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,6 +587,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "static_assertions",
  "thiserror",
 ]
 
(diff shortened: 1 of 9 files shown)


kornelski:
Do callbacks really need to be `Send + Sync`? This may be an annoying limitation, especially `Sync`.

jongiddy:
This causes problems with existing code if `Rc<RefCell<_>>` is used to share data between multiple closures. Such structures are required for keeping state over multiple closures, especially when using `on_end_tag` with its closure within a closure (e.g. https://github.com/jongiddy/strictly-data/blob/b5ff138149cfb0fd763fc1c44639c81388f5ebd2/src/extract.rs#L493-L499).

If this change is merged, would such structures need to be replaced with `Arc<Mutex<_>>` even if threading is not used?

jyn514:
> If this change is merged, would such structures need to be replaced with Arc<Mutex<_>> even if threading is not used?

It would, yes.

Ok, I think I'm convinced this isn't the right way to fix this.

PR #121, opened 2021-12-23
Add a C API for `on_end_tag`

jyn514:
- Expose `Element::on_end_tag`
- Expose all the methods of `EndTag` (`name`, `set_name`, `before`, `after`, `remove`)
- Add a new `EndTag::set_name_str` function to make sure it uses the proper encoding. Using
    `set_name` would require blindly assuming that the rewriter always used UTF8.
   This API can be slightly more efficient once https://github.com/hsivonen/encoding_rs/pull/77 is merged, but I'm not sure how long it will take, so it might be better to just merge this as-is and make a breaking change eventually.
- Add a test to ensure all the new API works correctly

Fixes https://github.com/cloudflare/lol-html/issues/85

### DIFF
--- a/c-api/include/lol_html.h
+++ b/c-api/include/lol_html.h
@@ -23,6 +23,7 @@ typedef struct lol_html_HtmlRewriterBuilder lol_html_rewriter_builder_t;
 typedef struct lol_html_HtmlRewriter lol_html_rewriter_t;
 typedef struct lol_html_Doctype lol_html_doctype_t;
 typedef struct lol_html_DocumentEnd lol_html_doc_end_t;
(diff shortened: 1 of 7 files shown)


kornelski:
BTW, regarding string efficiency problem, you can do:

```rust
match encode(&string) {
   Cow::Owned(changed) => changed,
   Cow::Borrowed(_) => string,
}
```

kornelski (approved):

jyn514:
> BTW, regarding string efficiency problem, you can do:

Oh, that's clever, thank you! I updated it to use that instead :)

PR #122, opened 2021-12-23
Add CI workflow

jyn514:
- Fix integration tests
- Fix most of the fuzzer scripts. I couldn't get AFL or honggfuzz working, and given that there are so many others I didn't bother working on it too long for now.
- Fix and deny clippy lints
- Test `cargo fmt`, `cargo clippy`, unit tests, integration tests, C API tests, and that fuzz crates build. In the future I'd like to also check that `Cargo.lock` doesn't change compared to the checked-in version, but I'm waiting to hear feedback on https://github.com/cloudflare/lol-html/pull/114 first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,20 +23,13 @@ jobs:
       - name: Get rust version
         id: rust-version
         run: echo "::set-output name=version::$(rustc --version)"
-      - name: Cache cargo index
+      - name: Cache cargo index and registry
         uses: actions/cache@v1
+        key: immutable # this will always be hit because it's a constant
         with:
-          path: ~/.cargo/registry/index
-          key: index-${{ runner.os }}-${{ github.run_number }}
-          restore-keys: |
-            index-${{ runner.os }}-
+          path: ~/.cargo/registry
       - name: Create lockfile
         run: cargo generate-lockfile
-      - name: Cache cargo registry
-        uses: actions/cache@v1
-        with:
-          path: ~/.cargo/registry/cache
-          key: registry-${{ runner.os }}-${{ steps.rust-version.outputs.version }}-${{ hashFiles('Cargo.lock') }}
       - name: Fetch dependencies
         run: cargo fetch
       - name: Cache target directory
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,58 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+env:
+  RUSTFLAGS: -Dwarnings
+  RUST_BACKTRACE: 1
+
+jobs:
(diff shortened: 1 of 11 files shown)


kornelski:
I'm not a fan of `rustfmt --check`, because it will reject well-formatted readable code for very pedantic reasons. Even tiny seemingly innocent changes like deletion of a line can cause rustfmt's heuristics to change and reject code it has formatted itself before. When this happens in CI, it makes contribution slow and PRs fail on irrelevant trivialities.

jyn514:
Hmm, if we don't enforce formatting it will definitely regress in the future. I'm not *super* tied to using rustfmt, but if someone has weird formatting it's nice to let them just run `cargo fmt` instead of having to fix it manually.

kornelski:
It's not ideal either way, but I suggest that you run `cargo fmt` from time to time or when you spot bad formatting. An extra space is not an catastrophic build-breaking event — unless you make it so.

jyn514:
Ok, I removed the rustfmt check :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install Rust
+        run: rustup update stable && rustup default stable
+      - name: Get rust version
+        id: rust-version
+        run: echo "::set-output name=version::$(rustc --version)"
+      - name: Cache cargo index
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry/index
+          key: index-${{ runner.os }}-${{ github.run_number }}
```
kornelski, line 30: registry is global and idempotent, so the cache key doesn't need to be per runner

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+          restore-keys: |
+            index-${{ runner.os }}-
+      - name: Create lockfile
+        run: cargo generate-lockfile
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry/cache
```
kornelski, line 38: crate tarballs are also global and immutable, not os specific

also this directory grows endlessly (old versions are never deleted). is there some cleanup/TTL for it?

kornelski (approved):

.github/workflows/ci.yml
jyn514: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB. If you exceed this limit, GitHub will save your cache but will begin evicting caches until the total size is less than 10 GB.

7 days seems like a reasonable TTL.

Good catch on the tarball being immutable :) I changed it.

PR #124, opened 2021-12-27
Take a `FnOnce` handler for end tags, not `FnMut`

jyn514:
lol-html already runs the handler at most once; this guarantees that won't change.

cc @harrishancock - this would make some optimizations for the workers runtime you mentioned possible.


kornelski (approved):

kornelski:
It's so nice to be able to just change the type and know it will be called at most once.

PR #128, opened 2022-03-19
Update html5lib-tests, install dependabot to open such PRs in the future

untitaker:


untitaker:
Ok, I fixed all tests I know how to fix. I am not sure about the remaining one as it seems lol-html doesn't implement any entity decoding outside of the testsuite, but entity decoding is supposed to happen for some text tokens but not others depending on context.

untitaker:
I believe this is ready for review, but I have some concerns about how lol-html's test harness interacts with tokenizer tests. A test like this would be entirely valid to check into html5lib-tests:

```
{"tests": [

{"description": "Truncated open tag after script open tag",
"input": "<script><c",
"output":[["StartTag", "script", {}]],
"errors": [
  {"code": "eof-in-tag", "line": 1, "col": 11}
]}

]}
```

...and lol-html seems to fail it as it will additionally emit `["Character", "<c"]`. That makes sense, after all lol-html doesn't want to modify the stream unnecessarily. But both parse5 and html5ever would pass this test.

untitaker:
@jyn514 any chance this can be reviewed? I am also curious on y'all's opinion about the last testcase I posted

src/parser/state_machine/syntax/comment.rs
```diff
@@ -9,15 +9,15 @@ define_state_group!(comment_states_group = {
     comment_start_state <-- ( create_comment; start_token_part; ) {
         b'-' => ( mark_comment_text_end; --> comment_start_dash_state )
         b'>' => ( mark_comment_text_end; emit_current_token?; --> data_state )
-        eof  => ( mark_comment_text_end; emit_current_token_and_eof?; )
+        eof  => ( reconsume in comment_state )
```
jyn514, line 12: Can you give an example of a test this fixes? Why is the change necessary? Ditto for the rest of the changes.

jyn514 (commented):
I am not super comfortable reviewing this part of the code :sweat_smile: @inikulin do you have opinions?

tests/harness/suites/html5lib_tests/test_token.rs
```diff
@@ -164,32 +164,42 @@ impl Unescape for TestToken {
...
+                    self.handled_text_decoding_until = decoded.len();
+                    self.tokens.push(TestToken::Text(decoded));
                 } else {
-                    self.0.push(TestToken::Text(if t.last_in_text_node() {
-                        decode_text(text, t.text_type())
-                    } else {
-                        text.into()
-                    }));
+                    self.handled_text_decoding_until = 0;
+                    self.tokens.push(TestToken::Text(text));
```
jyn514, line 194: This doesn't compile.
```
error[E0308]: mismatched types
   --> tests/harness/suites/html5lib_tests/test_token.rs:194:54
    |
194 |                     self.tokens.push(TestToken::Text(text));
    |                                                      ^^^^- help: try using a conversion method: `.to_string()`
    |                                                      |
    |                                                      expected struct `std::string::String`, found `&str`
```

src/parser/state_machine/syntax/comment.rs
untitaker: this is a direct update from the spec: https://html.spec.whatwg.org/multipage/parsing.html#comment-start-state

src/parser/state_machine/syntax/comment.rs
```diff
@@ -9,15 +9,15 @@ define_state_group!(comment_states_group = {
...
         _    => ( reconsume in comment_state )
     }
 
     comment_state {
         b'<' => ( --> comment_less_than_sign_state )
         b'-' => ( mark_comment_text_end; --> comment_end_dash_state )
         eof  => ( mark_comment_text_end; emit_current_token_and_eof?; )
-        _    => ()
+        _    => ( mark_comment_text_end; )
```
untitaker, line 20: https://html.spec.whatwg.org/multipage/parsing.html#comment-state

>Anything else: Append the [current input character](https://html.spec.whatwg.org/multipage/parsing.html#current-input-character) to the comment token's data.

src/parser/state_machine/syntax/comment.rs
```diff
@@ -42,16 +42,16 @@ define_state_group!(comment_states_group = {
     }
 
     comment_less_than_sign_state {
```
untitaker, line 44: https://html.spec.whatwg.org/multipage/parsing.html#comment-less-than-sign-state

both changed state transitions have "append the current input character to the comment token's data" in their actions

src/parser/state_machine/syntax/comment.rs
untitaker: currently playing around with the code to give you concrete testcases

jyn514: FWIW, if the spec says to do it, I'm ok with not having an outright test. I mostly wasn't sure why you'd made the change.

src/parser/state_machine/syntax/comment.rs
```diff
@@ -42,14 +42,14 @@ define_state_group!(comment_states_group = {
...
-        b'>' => ()
+        b'!' => ( mark_comment_text_end; --> comment_less_than_sign_bang_state )
+        b'<' => ( mark_comment_text_end; )
         eof  => ( reconsume in comment_state )
         _    => ( reconsume in comment_state )
     }
 
     comment_less_than_sign_bang_state {
-        b'-' => ( --> comment_less_than_sign_bang_dash_state )
+        b'-' => ( mark_comment_text_end; --> comment_less_than_sign_bang_dash_state )
```
untitaker, line 52: actually this is an interesting point I didn't fully figure out. it seems that `mark_comment_text_end` ends the current comment text at the _previous_ character. added a comment, hopefully it makes sense

src/parser/state_machine/syntax/comment.rs
untitaker: yes so I believe if you revert any single change here you should either:

1. no longer line up with the spec
2. break a testcase

I added some clarifying comments because mark_comment_text_end isn't the exact same action as "append current input character"

jyn514 (approved):
Sorry for the long delay. These changes all seem reasonable.

Issue #88, opened 2021-05-13
Removing the doctype


jed:
I'm not sure this is possible and I'm missing it, but it would be great to be able to remove the document's doctype.

gjtorikian:
Fancy running into you here, @jed ! 

I needed this for a project, too, and I just opened a PR for it: https://github.com/cloudflare/lol-html/pull/129

jyn514:
This seems reasonable to me, but I'm curious why you would need to remove the doctype?

jed:
In my case, the ability to remove the doctype means that you can stream HTML into valid JSON directly.

jed:
Very cool. Just out of curiosity, any idea when this would land in the production runtime?

jyn514:
@jed There is additional work on the runtime side to add this to JS; I'm not on the workers team and don't know whether they plan to add it at all.

jed:
Awesome, thanks. Let me know if there's anywhere I should help bring it up. :bow:

jyn514:
@evanderkoogh is involved in product on the workers side I think, he might know who to contact.

PR #129, opened 2022-05-18
Add ability to remove doctype

gjtorikian:
I needed this feature for a project I'm working on (a Ruby wrapper to this crate--I'll send a PR to update the README when it's done 😄 ).

Note that while I could've added `Mutations` to the `struct` like many of the other rewriteable tokens, this didn't seem appropriate for a couple of reasons:

* None of the other functions (like `before` or `after`) are necessary here
* `doctype` has its own `to_bytes` serialization, whereas the others use `impl_serialize!`

Closes https://github.com/cloudflare/lol-html/issues/88.

### DIFF
--- a/c-api/src/doctype.rs
+++ b/c-api/src/doctype.rs
@@ -24,3 +24,13 @@ pub extern "C" fn lol_html_doctype_user_data_set(doctype: *mut Doctype, user_dat
 pub extern "C" fn lol_html_doctype_user_data_get(doctype: *const Doctype) -> *mut c_void {
     get_user_data!(doctype)
 }
+
+#[no_mangle]
+pub extern "C" fn lol_html_doctype_remove(doctype: *mut Doctype) {
+    to_ref_mut!(doctype).remove();
+}
+
+#[no_mangle]
+pub extern "C" fn lol_html_doctype_is_removed(doctype: *const Doctype) -> bool {
+    to_ref!(doctype).removed()
+}
--- a/c-api/tests/src/test_doctype_api.c
+++ b/c-api/tests/src/test_doctype_api.c
@@ -112,7 +112,7 @@ static void test_get_user_data(void *user_data) {
 //-------------------------------------------------------------------------
 EXPECT_OUTPUT(
     remove_doctype_output_sink,
-    "<!DOCTYPE><html></html>",
+    "<html></html>",
     &EXPECTED_USER_DATA,
     sizeof(EXPECTED_USER_DATA)
 );
--- a/c-api/tests/src/test_doctype_api.c
+++ b/c-api/tests/src/test_doctype_api.c
@@ -110,6 +110,48 @@ static void test_get_user_data(void *user_data) {
 }
 
 //-------------------------------------------------------------------------
+EXPECT_OUTPUT(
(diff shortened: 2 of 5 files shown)


src/rewritable_units/tokens/doctype.rs
```diff
@@ -86,14 +89,28 @@ impl<'i> Doctype<'i> {
     pub fn force_quirks(&self) -> bool {
         self.force_quirks
     }
+
+    /// Removes the doctype.
+    #[inline]
+    pub fn remove(&mut self) {
+        self.removed = true
```
jyn514, line 96: No need to add a new field here.
```suggestion
        self.raw = Bytes::from_str("", Encoding::UTF8);
```

c-api/tests/src/test_doctype_api.c
```diff
@@ -109,6 +109,48 @@ static void test_get_user_data(void *user_data) {
     );
 }
 
+//-------------------------------------------------------------------------
+EXPECT_OUTPUT(
+    remove_doctype_output_sink,
+    "<!DOCTYPE><html></html>",
```
jyn514, line 115: shouldn't this have been removed?
```suggestion
    "<html></html>",
```

jyn514 (changes requested):

c-api/src/doctype.rs
```diff
@@ -24,3 +24,13 @@ pub extern "C" fn lol_html_doctype_user_data_set(doctype: *mut Doctype, user_dat
...
     get_user_data!(doctype)
 }
+
+#[no_mangle]
+pub extern "C" fn lol_html_doctype_remove(doctype: *mut Doctype) {
+    to_ref_mut!(doctype).remove();
+}
+
+#[no_mangle]
+pub extern "C" fn lol_html_doctype_is_removed(doctype: *const Doctype) -> bool {
```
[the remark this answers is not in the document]
gjtorikian: They do exist in other tags:

* https://github.com/cloudflare/lol-html/blob/2eed349dcdfa4ff5c19fe7c6e501cfd687601033/c-api/src/comment.rs#L58
* https://github.com/cloudflare/lol-html/blob/2eed349dcdfa4ff5c19fe7c6e501cfd687601033/c-api/src/element.rs#L207

src/rewritable_units/tokens/doctype.rs
gjtorikian: I can do this, but see above for the need to preserve the state as `bool` to match the behavior of other tags.

c-api/tests/src/test_doctype_api.c
gjtorikian: 😅 A copy pasta error!

c-api/src/doctype.rs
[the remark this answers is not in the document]
jyn514: Ah, thanks for the link. I agree it's good to have this for consistency.

src/rewritable_units/tokens/doctype.rs
jyn514: Gotcha, no need to change it.

jyn514 (approved):

PR #141, opened 2022-08-15
🐛 Patch CVE-2021-32810

imhunterand:
In the affected version of this crate, the result of the race condition is that one or more tasks in the worker queue can be popped twice instead of other tasks that are forgotten and never popped. If tasks are allocated on the heap, this can cause double free and a memory leak. If not, this still can cause a logical bug. Crates using `Stealer::steal`, `Stealer::steal_batch`, or `Stealer::steal_batch_and_pop` are affected by this issue.

**CVE-2021-32810**
`9.8/ 10`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,7 +140,7 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-deque"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


kornelski:
Two issues:

1. This crate doesn't use the `Stealer`, so this advisory doesn't seem relevant. Cargo doesn't use lockfiles for libraries, so downstream users aren't affected either way.

2. It seems that you've edited the lockfile by hand, because the checksum hasn't changed. This file needs to be maintained by Cargo.

I've updated the lockfile.

PR #144, opened 2022-08-16
Drop outdated pre-commit dependency

kornelski:
The `pre-commit` crate is unmaintained, uses prehistoric `rustc-serialize` that is even more dead, makes incorrect assumptions about Cargo workspaces. Plus modification of .git dir and sneaky modifications of commits are IMHO hacky and invasive.


jyn514 (approved):
Thanks for this, I've been meaning to ask about getting rid of it.

This leaves an isolated `scripts/precommit.sh` file - can you either document it in the readme or remove it altogether? No strong preference for which.

kornelski:
I'll drop it, since forced automatic reformatting is an option in most editors.

PR #146, opened 2022-10-15
Fix clippy lints

jongiddy:
Fix failing clippy lints, mostly adding `Eq` to `PartialEq` derives and removing Deref's that do not need to be explicit.

Also update `Cargo.lock` to match `Cargo.toml`.


jyn514 (approved):
Thanks!

PR #148, opened 2022-11-17
Added `Element::is_start_tag_self_closing()` and `Element::replace_start_tag()`.

orium:


src/rewritable_units/element.rs
```diff
@@ -127,6 +127,12 @@ impl<'r, 't> Element<'r, 't> {
         Ok(())
     }
 
+    /// Whether the starting tag of this element is explicitly self-closing, e.g. `<foo />`.
+    #[inline]
+    pub fn is_start_tag_self_closing(&self) -> bool {
```
jongiddy, line 132: This could just be called `is_self_closing` since it is a property of the element, not just the tag.

jongiddy (commented):
I think the example for `replace_start_tag` could be done using the existing methods.
```
el.remove_and_keep_content();
el.before("<div>", ContentType::Html);
el.after("</div>", ContentType::Html);
```

orium:
> I think the example for `replace_start_tag` could be done using the existing methods.

Yes, I think that works, but it's nice to have a more convenient method for this.

src/rewritable_units/element.rs
```diff
@@ -413,6 +419,37 @@ impl<'r, 't> Element<'r, 't> {
...
+    ///                 el.replace_start_tag("<div>");
+    ///
+    ///                 Ok(())
+    ///             })
+    ///         ],
+    ///         ..RewriteStrSettings::default()
+    ///     }
+    /// ).unwrap();
+    ///
+    /// assert_eq!(html, r#"<div>Hello</div>"#);
```
jyn514, line 446: can you please use the actual test case you care about (which I gather is something like `if el.is_self_closing() { el.replace_start_tag() }`) ?

jyn514 (commented):
> Edit: Actually, it is not that simple, because you need to handle the case of a both explicitly and implicitly self closing tags.

Can you explain this more? Why does handling both require the new `is_self_closing` method?

If you just want to remove all attributes, you can loop through https://docs.rs/lol_html/latest/lol_html/html_content/struct.Element.html#method.attributes and call `remove_attribute` on each.

src/rewritable_units/element.rs
orium: I replied in the other comment.  Basically my use case is not a common one and is hard for a reader to understand.  This is simple and to the point, which is what makes better documentation imo.

jongiddy:
For consistency with `EndTag` there could be a method on `Element` that returns the `StartTag` which then provides additional methods such as `remove`. The [example](https://docs.rs/lol_html/latest/lol_html/html_content/struct.Element.html#method.on_end_tag) for `on_end_tag` shows the same pattern for how to replace the tag.

This would support self-closing tags by providing a way to only remove the start tag.

orium:
> For consistency with EndTag there could be a method on Element that returns the StartTag which then provides additional methods such as remove. The [example](https://docs.rs/lol_html/latest/lol_html/html_content/struct.Element.html#method.on_end_tag) for on_end_tag shows the same pattern for how to replace the tag.

I like that.  Exposing the start tag will give more information to the client of the library (for instance, it exposes `StartTag::encoding()`).  I'm happy to replace `Element::replace_start_tag()` with just a method that returns a `&mut StartTag`.

I would say that it still makes sense to have `Element::is_self_closing()` because that being self closing is a property of the element as much as it is of the start tag.

orium:
I've removed in `Element::replace_start_tag()` favor of `Element::start_tag()`.  The start tag can then be modified via its `mutations` field:

```rust
element.start_tag().mutations.replace(⋯, ContentType::Html);
```

orium:
@jongiddy Can you take a look and see if it looks alright?

jongiddy:
`StartTag` should be public [here](https://github.com/cloudflare/lol-html/blob/1a1ab2e2bf896f815fe8888ed78ccdf46d7c6b85/src/lib.rs#L55).

I'm not sure if `Mutations` should be publicly accessible. It is `pub` in both `EndTag` and `StartTag`, but I'd prefer to promote its actions to methods on `StartTag`, so that `StartTag` had at least the methods from `EndTag`. Perhaps @inikulin  and @jyn514 have a better understanding of the conceptual model here?

orium:
> StartTag should be public [here](https://github.com/cloudflare/lol-html/blob/1a1ab2e2bf896f815fe8888ed78ccdf46d7c6b85/src/lib.rs#L55).

Done.

> I'm not sure if Mutations should be publicly accessible. It is pub in both EndTag and StartTag, but I'd prefer to promote its actions to methods on StartTag, so that StartTag has at least the [methods](https://docs.rs/lol_html/latest/lol_html/html_content/struct.EndTag.html) from EndTag. Perhaps @inikulin and @jyn514 have a better understanding of the conceptual model here?

I've added a few mutation methods to `StartTag`, bringing it on par with `EngTag`.  Also added a `replace()` to `EndTag`.

As for making `mutations` private: that's a breaking change (but I guess that's fine since we are v0.x), but it also doesn't seem to allow everything can be possible by exposing `Mutations`: in `Mutations` a bunch of things are are public and there can be people out there directly modifying those things.  We would have to ensure they have an alternative way to do so.

jongiddy (approved):
I think the `mutations` field on `StartTag` and `EndTag` should not be public. Either remove the `pub` or make it `pub(crate)` if that doesn't build.

`StartTag` was not previously public so this is a backwards-compatible change. For already-public `EndTag` it is technically backwards-incompatible, but I don't expect it will have been used. It's not documented and the `Mutations` type is not public. So I'm OK with dropping it there too.

Looks good with those changes.

jyn514 (approved):
Looks good, thanks!

PR #152, opened 2022-12-12
Added more maintainers.

orium:


Noah-Kennedy (approved):
Approving, but GH says it has errors. I think you got several other people's handles wrong, or they need to accept invites.

PR #161, opened 2023-03-08
Send + Sync implentation of HtmlRewriter

michaelvanstraten:
A few days ago, I came across this crate when I was researching methods of applying regular expression onto an HTML stream. I was really excited to see that this hard problem was actually solvable using this crate. 
I sadly realized that the [`HtmlRewriter`](https://docs.rs/lol_html/0.3.2/lol_html/struct.HtmlRewriter.html) is nether Send nor Sync, so it is almost impossible to us it in combination with futures. 

So I decided to try to do a basic rewrite of the crate into a Send and Sync implementation. It currently compiles and all the tests pass. 

Although I refrained from using any RwLock's or Mutex's, there are some minor/major performance and throughput regressions, I attached an output file of a run I did using the script/bench.sh benchmark. 
[benchmark.txt](https://github.com/cloudflare/lol-html/files/10922950/benchmark.txt)

This and the bounds required for the [`HtmlRewriter`](https://docs.rs/lol_html/0.3.2/lol_html/struct.HtmlRewriter.html)  to be Send and Sync make it probably moronic to merge it directly into the lol-html crate. 
But it would be possible to put it behind a feature flag with almost no downsides for developers and users. 

This is the reason for the Pull request, do you guys think a Send and Sync version of the [`HtmlRewriter`](https://docs.rs/lol_html/0.3.2/lol_html/struct.HtmlRewriter.html) behind a feature flag would be mergeable? 
I ask because I would have to put in some effort and time to not just duplicate some code a bunch of time (Send and non Send versions).

Thank you very much in advance, I am looking forward to hearing back from you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ safemem = "0.3.3"
 selectors = "0.22.0"
 thiserror = "1.0.2"
 hashbrown = "0.13.1"
+atomic_refcell = "0.1.9"
 
 [dev-dependencies]
 criterion = "0.4.0"
--- a/src/memory/arena.rs
+++ b/src/memory/arena.rs
@@ -63,12 +63,12 @@ impl Arena {
 mod tests {
     use super::super::limiter::MemoryLimiter;
     use super::*;
-    use std::rc::Rc;
+    use std::sync::Arc;
 
     #[test]
     fn append() {
         let limiter = MemoryLimiter::new_shared(10);
-        let mut arena = Arena::new(Rc::clone(&limiter), 2);
+        let mut arena = Arena::new(Arc::clone(&limiter), 2);
 
         arena.append(&[1, 2]).unwrap();
         assert_eq!(arena.bytes(), &[1, 2]);
@@ -90,7 +90,7 @@ mod tests {
     #[test]
     fn init_with() {
         let limiter = MemoryLimiter::new_shared(5);
-        let mut arena = Arena::new(Rc::clone(&limiter), 0);
+        let mut arena = Arena::new(Arc::clone(&limiter), 0);
 
         arena.init_with(&[1]).unwrap();
         assert_eq!(arena.bytes(), &[1]);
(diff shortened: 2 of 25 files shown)


michaelvanstraten:
@inikulin?

inikulin:
Thank you for the contribution.

The rewriter is designed for single-threaded uses. To be used in multithreaded environment and satisfy `Send + Sync` it just needs external synchronisation  (i.e. be wrapped in `Mutex` or `RwLock`).

michaelvanstraten:
Could you please reopen this PR?

It is impossible to `Send`any type that it not send, wrapping it in a `Arc<Mutex>` doesn't make it Send. 
Please check the respective docs for [`Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#impl-Send-for-Mutex%3CT%3E) as well as [`Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send-for-Arc%3CT%3E).

`Send` and `Sync` is only ever implemented for a type if all the contained types in it also implement `Send` and `Sync`. This makes sense, so to ensure that things that are not `Send` can't be sent across threads. For example: A Thread Local only ever makes sense to use if it is used from the same thread, wrapping it in a `Arc<Mutex>` should not make it sendable.

inikulin:
Indeed, my bad. 

However, unfortunately, due to performance considerations we won't accept this PR. Even though compiler marks `HtmlRewriter` as `!Send` because of `Rc`s, those `Rc`s are never leaked outside of the rewriter, so it's safe to assume that `HtmlRewriter` is `Send` as long as its handlers are `Send` and its behind `Mutex`. So, as a workaround you can do:
```rust
struct SendRewriter(Mutex<HtmlRewriter>);

unsafe impl Send for SendRewriter {}
``` 
+ add wrapper methods that statically ensure that rewriter handlers are `Send`.

dxvid-pts:
Maybe using https://crates.io/crates/sendable (which implements send for Rc) does not affect performance such as using Arc. Interested in the send implementation as well.

PR #162, opened 2023-03-14
Support dynamic charset change on `meta` tags in `HtmlRewriter`.

orium:


src/rewriter/mod.rs
```diff
@@ -12,13 +12,46 @@ use crate::parser::ParsingAmbiguityError;
...
 use std::error::Error as StdError;
 use std::fmt::{self, Debug};
+use std::ops::Deref;
 use std::rc::Rc;
 use thiserror::Error;
 
 pub use self::settings::*;
 
+#[derive(Clone)]
+pub struct SharedEncoding {
```
inikulin, line 27: Move to `base` dir

src/rewriter/mod.rs
```diff
@@ -140,6 +173,40 @@ macro_rules! guarded {
     }};
 }
 
+fn ascii_compatible_encoding_from_mimetype(mime: &Mime) -> Option<AsciiCompatibleEncoding> {
```
inikulin, line 176: make a method on `AsciiCompatibleEncoding`

src/rewriter/mod.rs
```diff
@@ -140,6 +173,40 @@ macro_rules! guarded {
...
+    mime.get_param("charset")
+        .and_then(|cs| Encoding::for_label_no_replacement(cs.as_str().as_bytes()))
+        .and_then(AsciiCompatibleEncoding::new)
+}
+
+fn handler_adjust_charset_on_meta_tag(
```
inikulin, line 182: Move method after `impl HtmlRewriter {}`

src/rewriter/mod.rs
```diff
@@ -150,12 +217,24 @@ impl<'h, O: OutputSink> HtmlRewriter<'h, O> {
...
-        let encoding = settings.encoding;
+        let encoding = SharedEncoding::new(settings.encoding);
         let mut selectors_ast = selectors_vm::Ast::default();
         let mut dispatcher = ContentHandlersDispatcher::default();
-        let has_selectors = !settings.element_content_handlers.is_empty();
+        let has_selectors =
+            !settings.element_content_handlers.is_empty() || settings.adjust_charset_on_meta_tag;
+
+        let charset_adjust_handler = match settings.adjust_charset_on_meta_tag {
+            true => Some(handler_adjust_charset_on_meta_tag(SharedEncoding::clone(
```
inikulin, line 227: Just use `if .. else`

src/rewriter/mod.rs
```diff
@@ -558,6 +649,110 @@ mod tests {
...
+        use crate::html_content::{ContentType, TextChunk};
+
+        let enthusiastic_text_handler = || {
+            doc_text!(move |text: &mut TextChunk| {
+                let new_text = text.as_str().replace('!', "!!!");
+                text.replace(&new_text, ContentType::Text);
+                Ok(())
+            })
+        };
+        let html: Vec<u8> = [
```
inikulin, line 663: Here and in all other places: separate variable declarations, multiline expressions and statements by empty lines

src/rewriter/mod.rs
```diff
@@ -558,6 +649,110 @@ mod tests {
...
+            r#"<meta http-equiv="content-type" content="text/html; charset=windows-1251"><html><head></head><body>I love "#.as_bytes().to_vec(),
+            vec![0xd5, 0xec, 0xb3, 0xcb, 0xdc],
+            r#"!</body></html>"#.as_bytes().to_vec(),
+        ].into_iter().concat();
```
inikulin, line 721: This needs `cargo fmt`

tests/fixtures/token_capturing.rs
```diff
@@ -2,10 +2,7 @@ use crate::harness::suites::html5lib_tests::{
     get_test_cases, TestCase, TestToken, TestTokenList,
 };
 use crate::harness::{TestFixture, Input};
-use lol_html::{
-    LocalNameHash, TokenCaptureFlags, LocalName, Token, StartTagHandlingResult, TransformController,
-    TransformStream, Namespace, TransformStreamSettings, MemoryLimiter
-};
+use lol_html::{LocalNameHash, TokenCaptureFlags, LocalName, Token, StartTagHandlingResult, TransformController, TransformStream, Namespace, TransformStreamSettings, MemoryLimiter, SharedEncoding};
```
inikulin, line 5: This needs cargo fmt. Let's add `cargo fmt -- --check` to CI

inikulin (changes requested):

src/rewriter/mod.rs
orium: I find matching on booleans to often be more readable than a `if`, specially when both expressions are similar (not the case here).

Happy to change it anyway.

src/rewriter/mod.rs
orium: This is formatted by `cargo fmt`.

tests/fixtures/token_capturing.rs
orium: This is formatted by `cargo fmt`.

> Let's add cargo fmt -- --check to CI

Will open a PR for that.

src/rewriter/mod.rs
orium: Done.  Does it make sense to also move `AsciiCompatibleEncoding` to `base/encoding.rs`?  We would export it in a different place so it is a breaking change...

orium: Actually, `SharedEncoding` is not exported.  It is only used internally.  Do you still want to have it in the `base` module?

src/rewriter/mod.rs
orium: :heavy_check_mark:

src/rewriter/mod.rs
orium: :heavy_check_mark:

src/rewriter/mod.rs
orium: :heavy_check_mark:

src/rewriter/mod.rs
orium: :heavy_check_mark:

inikulin (approved):

Issue #160, opened 2023-03-02
Detecting void elements


mwcz:
Thanks for lol-html; it's delightful!

I'm in a situation where I need to detect self-closing tags, including [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements).  `Element::is_self_closing` detects explicit self-closing tags (`<foo />`), but not void elements, which are implicitly self-closing (`<img>`).  AFAICT, whether an element is void is known by stack.rs, but is later mixed into `element.can_have_content` and the exact voidness value is lost.

Would any of the following options be palatable?

 1. Change `Element::is_self_closing` to return `true` for explicit self-closing tags _and_ void elements (ie, have it return `!self.can_have_content` instead of `self.start_tag.self_closing()`).
 2. Add a new function, like `Element::is_void` which bubbles up the result of stack.rs's `is_void_element`.  This seems like it would require the most code changes.
 3. Add a new function, like `Element::can_have_content` which returns `self.can_have_content`.  One or the other can be renamed to avoid the name collision.
 4. Some other option? :thinking: 

Any one of these would be enough to do what I'm hoping to do, and I'm happy to put together a PR after a tip on what the best approach is.  Thanks!

Jared-Sprague:
Yes we really need this!  Would love to see this fixed.

mwcz:
I added a draft PR the patch we're using in the meantime.

orium:
Closed by #163.

PR #163, opened 2023-03-14
Add `can_have_content` getter for Element

mwcz:
Here's an example implementation of item 3 from #160.

### DIFF
--- a/src/rewritable_units/element.rs
+++ b/src/rewritable_units/element.rs
@@ -133,6 +133,14 @@ impl<'r, 't> Element<'r, 't> {
         self.start_tag.self_closing()
     }
 
+    /// Whether the element can have inner content.  Returns `true` unless the element is an [HTML void
+    /// element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements) or has a
+    /// self-closing tag (eg, `<foo />`).
+    #[inline]
+    pub fn can_have_content(&self) -> bool {
+        self.can_have_content
+    }
+
     /// Returns the [namespace URI] of the element.
     ///
     /// [namespace URI]: https://developer.mozilla.org/en-US/docs/Web/API/Element/namespaceURI


src/rewritable_units/element.rs
```diff
@@ -46,22 +46,22 @@ pub struct Element<'r, 't> {
     end_tag_mutations: Option<Mutations>,
     modified_end_tag_name: Option<Bytes<'static>>,
     end_tag_handler: Option<EndTagHandler<'static>>,
-    can_have_content: bool,
+    el_can_have_content: bool,
```
orium, line 49: Given that this is a `Element` the `el_` doesn't add much.  I think the original name is better.

src/rewritable_units/element.rs
```diff
@@ -133,6 +133,14 @@ impl<'r, 't> Element<'r, 't> {
         self.start_tag.self_closing()
     }
 
+    /// Whether the element can have inner content.  In other words, is it either a self-closing
```
orium, line 136: > In other words, is it either a self-closing tag like `<foo />` or is an HTML void element which are implicitly self-closing, e.g. ``.

This is the situation when the method returns false.  Can you clarify by saying something like "This returns `true` unless it is a self-closing tag ..." ?

src/rewritable_units/element.rs
mwcz: I agree, but the field and the function can't have the same name.  Maybe the field could keep it's name and we just make it `pub` instead?

orium: > the field and the function can't have the same name

It can.  Give it a try :P

mwcz: Weird, I could have sworn I tried that first and the compiler didn't like it.  Fixed.

src/rewritable_units/element.rs
mwcz: Changed to:

```
Whether the element can have inner content.  Returns `true` unless the element is an [HTML void
element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements) or has a
self-closing tag (eg, `<foo />`).
```

src/rewritable_units/element.rs
```diff
@@ -53,15 +53,15 @@ pub struct Element<'r, 't> {
 }
 
 impl<'r, 't> Element<'r, 't> {
-    pub(crate) fn new(start_tag: &'r mut StartTag<'t>, can_have_content: bool) -> Self {
+    pub(crate) fn new(start_tag: &'r mut StartTag<'t>, el_can_have_content: bool) -> Self {
```
orium, line 56: You can also remove the `el_` here.

mwcz: Good catch, LSP rename missed that one, as did I.

orium:
Needs to land after https://github.com/cloudflare/lol-html/pull/165 since it fixes some warnings.

orium:
Just landed #165 which should fix the warnings.  Can you rebase on master and squash your commits into a single one please?

orium (approved):

orium:
Merged.  Thank you!

PR #175, opened 2023-04-20
Added methods to do in-place modifications to the text in a `TextChunk`.

orium:


src/rewritable_units/tokens/text_chunk.rs
```diff
@@ -89,6 +89,18 @@ impl<'i> TextChunk<'i> {
...
 
+    /// Sets the textual content of the chunk.
+    #[inline]
+    pub fn set_str(&mut self, text: String) {
+        self.text = Cow::Owned(text);
+    }
+
+    /// WIP!
+    #[inline]
+    pub fn modify_str(&mut self, mut f: impl FnMut(&mut String)) {
```
orium, line 100: Alternatively I can just have a `pub fn as_mut_str(&mut self) -> &mut String { self.text.to_mut() }

inikulin (approved):

PR #176, opened 2023-04-22
Fix `afl` and `honggfuzz` fuzzing scripts

Kuribohrn:
This commit fixes the broken `afl` fuzzing script by adding  `lolhtml` library at the linking stage and runtime.


kornelski:
I think the linking has already been fixed a00705aa40f96fa73b18cfb8a758c62729224258

For afl the crate should be `cargo-afl`

PR #187, opened 2023-06-07
Fix lol_html_take_last_error() after calling lol_html_element_add_end_tag_handler() on element with no end tag

harrishancock:
PR #182 included a tiny regression: https://github.com/cloudflare/lol-html/pull/182#discussion_r1221337721

I hacked a little fix together. Suggestions for greater elegance are welcome. :)

### DIFF
--- a/c-api/src/element.rs
+++ b/c-api/src/element.rs
@@ -228,18 +228,18 @@ pub extern "C" fn lol_html_element_add_end_tag_handler(
 ) -> c_int {
     let element = to_ref_mut!(element);
 
-    match element.end_tag_handlers() {
-        Some(handlers) => {
-            handlers.push(Box::new(move |end_tag| {
-                match unsafe { handler(end_tag, user_data) } {
-                    RewriterDirective::Continue => Ok(()),
-                    RewriterDirective::Stop => Err("The rewriter has been stopped.".into()),
-                }
-            }));
-            0
+    let handlers = unwrap_or_ret_err_code! {
+        element.end_tag_handlers().ok_or("No end tag.")
+    };
+
+    handlers.push(Box::new(move |end_tag| {
+        match unsafe { handler(end_tag, user_data) } {
+            RewriterDirective::Continue => Ok(()),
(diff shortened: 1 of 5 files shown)


c-api/src/element.rs
```diff
@@ -228,18 +228,22 @@ pub extern "C" fn lol_html_element_add_end_tag_handler(
...
+            Some(handlers) => {
+                handlers.push(Box::new(move |end_tag| {
+                    match unsafe { handler(end_tag, user_data) } {
+                        RewriterDirective::Continue => Ok(()),
+                        RewriterDirective::Stop => Err("The rewriter has been stopped.".into()),
+                    }
+                }));
+                Ok::<(), &'static str>(())
+            }
+            None => Err("No end tag.".into())
```
inikulin, line 242: Nit, we could simplify it a bit:

```rust
let handlers = unwrap_or_ret_err_code! { 
     element.end_tag_handlers().ok_or_else(|| "No end tag.".into()))
};

handlers.push(...);

0
```

harrishancock: Thanks, that's much cleaner.

`ok_or_else(|| "No end tag.".into())` causes the following error:
```
error[E0282]: type annotations needed
   --> /home/harris/github/lol-html/c-api/src/element.rs:232:64
    |
232 |         element.end_tag_handlers().ok_or_else(|| "No end tag.".into())
    |                                                                ^^^^
    |
   ::: /home/harris/github/lol-html/c-api/src/lib.rs:67:75
    |
67  |                 crate::errors::LAST_ERROR.with(|e| *e.borrow_mut() = Some(err.into()));
    |                                                                           --- type must be known at this point
    |
help: try using a fully qualified path to specify the expected types
    |
232 |         element.end_tag_handlers().ok_or_else(|| <&str as Into<T>>::into("No end tag."))
    |                                                  ++++++++++++++++++++++++             ~

```
Without the `.into()`, everything builds and tests pass, so I'm going to go with that unless that's a mistake for some reason.

inikulin (approved):

PR #190, opened 2023-07-25
Expose tag and attribute names with the original case.

orium:


scotchmist (approved):

kflansburg (approved):
This seems safe enough since it is not modifying any existing methods. Is this something we want to expose via the C API?

orium:
> This seems safe enough since it is not modifying any existing methods. Is this something we want to expose via the C API?

Sounds reasonable.  Added.

orium:
I've added a release commit to streamline things.

inikulin (approved):

PR #194, opened 2023-09-10
Make `ElementContentHandlers` and `DocumentContentHandlers` fields public.

orium:
This is helpful if you want to patch already created handlers with extra functionallity. For instance, if you want to measure the time a handler takes to run you can do something like:

```rust
fn measure_element_time_handler(handlers: ElementContentHandlers<'_>) -> = ElementContentHandlers<'_> {
   let element_handler = self.element.take().map(|handler| {
       |element| {
           // measure start time here.
           let result = handler(element);
           // measure end time here.
           result
       }
   });

   ElementContentHandlers {
       element: element_handler,
       comments: self.comments.take(),
       text: self.text.take(),
   }
}
```


scotchmist (approved):

src/rewriter/settings.rs
```diff
@@ -5,7 +5,7 @@ use super::AsciiCompatibleEncoding;
 use std::borrow::Cow;
 use std::error::Error;
 
-pub(crate) type HandlerResult = Result<(), Box<dyn Error + Send + Sync>>;
+pub type HandlerResult = Result<(), Box<dyn Error + Send + Sync>>;
```
inikulin, line 8: We need doc comments for those and all the newly exposed API. Let's add `#![warn(missing_docs)]` to the crate's lib.rs and also use `RUSTFLAGS="-D warnings" to the CI builds (if we don't have it already)  to fail on missing docs

inikulin (changes requested):

src/rewriter/settings.rs
orium: Will add documentation to those fields.  However, we can't fail on missing documentation right because it is missing from a lot of places:

```
warning: `lol_html` (lib) generated 35 warnings
```

inikulin: Let's revisit those and suppress the warning where it's intentional or provide docs otherwise.

inikulin (approved):

PR #217, opened 2024-08-06
WIP Make `HtmlRewriter` `Send + Sync`.

orium:


orium:
The performance of the rewriter is about 60% slower if we make it `Send + Sync`. For that reason I'll drop the PR.

PR #219, opened 2024-08-16
Avoid using a `Rc<RefCell<_>>` in `HtmlRewriteController::handlers_dispatcher()`.

orium:
This means one less lock in a potential `Send + Sync` version of the rewriter.


kornelski (approved):
edition 2021 closure capture rules make it simple.

Noah-Kennedy (approved):

PR #220, opened 2024-08-22
Removed RefCells for dispatcher.

orium:
We now explicitly pass that state around. This will reduce the number of `Mutex`es in a `Send` version of the rewriter.

Benchmarks show a median speedup of 0.4%... so, no performance difference when all benchmarks are aggregated.

### DIFF
--- a/src/parser/state_machine/mod.rs
+++ b/src/parser/state_machine/mod.rs
@@ -72,46 +72,62 @@ pub type StateResult = Result<(), ParsingTermination>;
 pub type ParseResult = Result<Never, ParsingTermination>;
 
 pub trait StateMachineActions {
-    fn emit_eof(&mut self, input: &[u8]) -> ActionResult;
-    fn emit_text(&mut self, input: &[u8]) -> ActionResult;
-    fn emit_current_token(&mut self, input: &[u8]) -> ActionResult;
-    fn emit_tag(&mut self, input: &[u8]) -> ActionResult;
-    fn emit_current_token_and_eof(&mut self, input: &[u8]) -> ActionResult;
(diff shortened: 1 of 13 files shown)


src/parser/state_machine/mod.rs
```diff
@@ -72,46 +72,73 @@ pub type StateResult = Result<(), ParsingTermination>;
...
-
-    fn set_closing_quote_to_double(&mut self, input: &[u8]);
-    fn set_closing_quote_to_single(&mut self, input: &[u8]);
-
-    fn mark_tag_start(&mut self, input: &[u8]);
-    fn unmark_tag_start(&mut self, input: &[u8]);
-
-    fn enter_cdata(&mut self, input: &[u8]);
-    fn leave_cdata(&mut self, input: &[u8]);
+    type PassingState;
```
orium, line 75: Looking for a better name for this.

scotchmist: I know it's overused as a term, but Context?

orium: Yeah, I think that's good: it is generic, but the nature of this type is also generic!

inikulin: yeah, was about to write that it's better to use `ParserContext`. Also, drop `passing` and `shared` prefixes. Just pass `ctx: ParserContext` around

orium: >why it needs to be a type parameter? is it because the state is generic over S.

It's because we need to reference it in the trait methods. (The ones in directly in the trait and the ones added by macros on the `StateMachine` "sub"-trait).

> Can we just make StateMachine trait generic then?

Not sure if I understand: the traits are generic over the context.

Noah-Kennedy (approved):

PR #221, opened 2024-08-22
Use internal mutability for `SharedMemoryLimiter`, removing one `RefCell`.

orium:
In a `Send` version the `Cell<usize>` can become an `AtomicUsize`.


Noah-Kennedy (approved):
nice!

PR #222, opened 2024-08-26
Attribute buffer is no longer shared, removing one `RefCell`.

orium:
~This makes the median benchmark 3.6% slower.~ I'm now not pre-allocating the vector for the attributes and the median benchmark is now only 3.1% slower.

### DIFF
--- a/src/parser/lexer/actions.rs
+++ b/src/parser/lexer/actions.rs
@@ -136,13 +136,11 @@ impl<S: LexemeSink> StateMachineActions for Lexer<S> {
 
     #[inline]
     fn create_start_tag(&mut self, _context: &mut ParserContext<S>, _input: &[u8]) {
-        self.attr_buffer.borrow_mut().clear();
-
         self.current_tag_token = Some(StartTag {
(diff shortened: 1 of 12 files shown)


src/parser/lexer/actions.rs
```diff
@@ -122,13 +122,11 @@ impl<S: LexemeSink> StateMachineActions for Lexer<S> {
...
     #[inline]
     fn create_start_tag(&mut self, _input: &[u8]) {
-        self.attr_buffer.borrow_mut().clear();
-
         self.current_tag_token = Some(StartTag {
             name: Range::default(),
             name_hash: LocalNameHash::new(),
             ns: Namespace::default(),
-            attributes: Rc::clone(&self.attr_buffer),
+            attributes: Vec::with_capacity(DEFAULT_ATTR_BUFFER_CAPACITY),
```
orium, line 129: I think we can just create the attributes here instead of reusing the buffer (which IIUC was an optimization).

inikulin (approved):

PR #223, opened 2024-09-04
[For review and discussion] Html rewriter send

orium:
The first commit is the PR #222.

The median performance change from the `Send` version to PR #222 is a slowdown of 1.7%.


orium:
Superseded by #224.

PR #224, opened 2024-09-10
Send parameterizable rewriter

orium:
Some data structures had to be turned into `Sync` to make the rewriter `Send` (if the handlers are `Send`). This has a performance penalty even when using a non-`Send` rewriter. 

The benchmarks have a median slowdown of  -0.8%. You read that right: a negative slowdown, which is a speedup. This  can only be explained by noise while running the benchmarks, as they can only get slightly slower, not faster. However, it shows that the performance change is negligible (in my x86 machine, at least).

### DIFF
--- a/benches/cases/rewriting.rs
+++ b/benches/cases/rewriting.rs
@@ -13,7 +13,7 @@ define_group!(
 
                     Ok(())
                 })],
-                ..Settings::default()
+                ..Settings::new()
             }
         ),
         (
@@ -24,7 +24,7 @@ define_group!(
 
                     Ok(())
                 })],
-                ..Settings::default()
+                ..Settings::new()
             }
         )
     ]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@
 //!
 //! The crate serves as a back-end for the HTML rewriting functionality of [Cloudflare Workers], but
 //! can be used as a standalone library with the convenient API for a wide variety of HTML
-//! rewriting/analyzis tasks.
+//! rewriting/analysis tasks.
 //!
 //! The crate provides two main API entry points:
 //!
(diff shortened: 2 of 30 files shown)


src/rewriter/settings.rs
```diff
@@ -5,8 +5,109 @@ use super::AsciiCompatibleEncoding;
...
+    // possible version of a handler (i.e. if we have `Send` and non-`Send` handlers we need to
+    // create a `Send` handler since that's compatible with both classes of handlers), so that's
+    // what we offer below.
+    //
+    // Note that in the HTML rewriter all we have is an abstract `H` that implements `HandlerTypes`.
+    // Therefore, there is no direct way of create a handler that is compatible with *all* possible
+    // implementations of `HandlerTypes`, so each implementation of `HandlerTypes` needs to prove
+    // that a handler compatible with itself is creatable.
+
+    fn new_end_tag_handler<'h>(
```
orium, line 29: I'm still working at improving these methods.

src/rewriter/mod.rs
```diff
@@ -371,9 +380,41 @@ mod tests {
         out
     }
 
+    #[allow(clippy::drop_non_drop)]
+    #[test]
+    fn handlers_covariance() {
```
inikulin, line 385: What's the purpose of this test? It doesn't seem to have any assertions

benches/cases/rewriting.rs
```diff
@@ -13,7 +13,7 @@ define_group!(
 
                     Ok(())
                 })],
-                ..Settings::default()
+                ..Settings::new()
```
inikulin, line 16: Why changing this to `new`?

src/lib.rs
```diff
@@ -37,12 +37,49 @@ use cfg_if::cfg_if;
 
 pub use self::rewriter::{
     rewrite_str, AsciiCompatibleEncoding, CommentHandler, DoctypeHandler, DocumentContentHandlers,
-    ElementContentHandlers, ElementHandler, EndHandler, EndTagHandler, HandlerResult, HtmlRewriter,
-    MemorySettings, RewriteStrSettings, Settings, TextHandler,
+    ElementContentHandlers, ElementHandler, EndHandler, EndTagHandler, HandlerNormalTypes,
```
inikulin, line 40: Let's rename `HandlerSendTypes` and `HanderNormalTypes` to `SendHandlerTypes` and `LocalHandlerTypes`

src/rewriter/settings.rs
```diff
@@ -5,8 +5,118 @@ use super::AsciiCompatibleEncoding;
...
+        handler: impl IntoHandler<EndTagHandlerSend<'h>>,
+    ) -> Self::EndTagHandler<'h>;
+
+    /// Creates an element handler.
+    fn new_element_handler<'h>(
+        handler: impl IntoHandler<ElementHandlerSend<'h, Self>>,
+    ) -> Self::ElementHandler<'h>;
+
+    /// Creates a handler by running multiple handlers in sequence.
+    fn combine_handlers(handlers: Vec<Self::EndTagHandler<'_>>) -> Self::EndTagHandler<'_>;
```
inikulin, line 47: This seems to be used only internally, should we use `#[doc(hidden)]`?

src/selectors_vm/mod.rs
```diff
@@ -143,7 +144,10 @@ pub struct SelectorMatchingVm<E: ElementData> {
     enable_esi_tags: bool,
 }
 
-impl<E: ElementData> SelectorMatchingVm<E> {
+impl<E: ElementData> SelectorMatchingVm<E>
+where
+    E: Send,
```
inikulin, line 149: Nit: move `ElementData` to the `where` clause

src/rewriter/mod.rs
orium: It checks the the handlers continue to be covariant, that is, if you have a handler with a lifetime larger than `'a` you can put that handler in a place where a lifetime of `'a` is expected.

The test doesn't assert anything because it is a "if it compile it is good". If the handlers stopped being covariant in their lifetime it test would not compile.

I'll add some comments.

benches/cases/rewriting.rs
orium: `Settings::new()` creates settings for a non-`Send` rewriter (that is, it creates a `Settings<'_, '_, HandlerNormalTypes>`).

`Settings::new_send()` creates settings for a `Send` rewriter (that is, it creates a `Settings<'_, '_, HandlerSendTypes>`).

`Settings::default()` also creates settings for a non-`Send` rewriter, just like `Settings::new()`.

The reason I added `Settings::new()` is because I had to add `Settings::new_send()` so it would make sense to also have `Settings::new()` as well.

PR #228, opened 2024-10-28
Minor tweaks and fixes

kornelski:


.rustfmt.toml
```diff
@@ -0,0 +1 @@
+
```
orium, line 1: Accidentally added?

src/selectors_vm/parser.rs
```diff
@@ -33,7 +33,7 @@ pub enum PseudoElementStub {}
 
 impl ToCss for PseudoElementStub {
     fn to_css<W: fmt::Write>(&self, _dest: &mut W) -> fmt::Result {
-        match *self {}
+        unimplemented!()
```
orium, line 36: Can you revert these `unimplemented!()`

Before these were proving that it was impossible to panic. Now it looks like they can panic (of of course they can't because this is unreachable), and it looks like it's unimplemented when that's not the case.

.rustfmt.toml
kornelski: Intentional. Otherwise my custom config is applied, and completely changes formatting of this project.

src/selectors_vm/parser.rs
kornelski: Previous code triggers a warning that it is UB to dereference an uninhabited type.

orium: Oh, interesting. In that case can we make it `unreachable!()` that represents what is it a bit better.

.rustfmt.toml
orium: Ok. Can you add a comment there that says the intention is to override system-wide rustfmt config?

orium (approved):

PR #229, opened 2024-10-28
Streaming handlers

kornelski:
Basis for https://github.com/cloudflare/workerd/issues/2758

* `to_bytes` turned into fallible `into_bytes` to guarantee that the content handlers are called only once, and to report encoding errors in the future.
* `into_bytes` using `self` became more performance-sensitive to the size of the token structs, so I've made `Mutations` allocated lazily, assuming that majority of nodes aren't mutated.
* there's a fast path that avoids using `encoding_rs::Encoder` when the output is UTF-8.
* This PR omits [support for incomplete UTF-8 chunks](https://github.com/cloudflare/lol-html/pull/241), I'll add that after this is merged.

### DIFF
--- a/src/html/mod.rs
+++ b/src/html/mod.rs
@@ -1,3 +1,6 @@
+use crate::base::Bytes;
+use memchr::{memchr, memchr3};
+
 #[macro_use]
 mod tag;
 
@@ -9,3 +12,63 @@ pub use self::local_name::{LocalName, LocalNameHash};
 pub use self::namespace::Namespace;
 pub use self::tag::Tag;
 pub use self::text_type::TextType;
+
+/// Convert text to HTML
+#[inline]
+pub(crate) fn escape_body_text(mut content: &str, output_handler: &mut impl FnMut(&str)) {
+    loop {
+        if let Some(pos) = memchr3(b'&', b'<', b'>', content.as_bytes()) {
+            let Some((chunk_before, (matched, rest))) = content
+                .split_at_checked(pos)
+                .and_then(|(before, rest)| Some((before, rest.split_at_checked(1)?)))
+            else {
+                return;
+            };
+            content = rest;
+            let matched = matched.as_bytes()[0];
+
+            if !chunk_before.is_empty() {
+                (output_handler)(chunk_before);
+            }
+            (output_handler)(match matched {
+                b'<' => "&lt;",
+                b'>' => "&gt;",
+                _ => "&amp;",
+            });
+        } else {
+            if !content.is_empty() {
+                (output_handler)(content);
+            }
+            return;
+        }
+    }
+}
+
+/// Replace `"` with `&quot;` ONLY, leaving `&` unescaped
+pub(crate) fn escape_double_quotes_only(
+    content: &Bytes<'_>,
+    output_handler: &mut dyn FnMut(&[u8]),
+) {
+    let mut content = &**content;
+    loop {
+        if let Some(pos) = memchr(b'"', content) {
+            let Some((chunk_before, rest)) = content
+                .split_at_checked(pos)
+                .and_then(|(before, rest)| Some((before, rest.get(1..)?)))
+            else {
+                return;
+            };
+            content = rest;
+
+            if !chunk_before.is_empty() {
+                (output_handler)(chunk_before);
+            }
+            (output_handler)(b"&quot;");
+        } else {
+            if !content.is_empty() {
+                (output_handler)(content);
+            }
+            return;
+        }
+    }
+}
--- a/src/html/mod.rs
+++ b/src/html/mod.rs
@@ -18,12 +18,13 @@ pub use self::text_type::TextType;
 pub(crate) fn escape_body_text(mut content: &str, output_handler: &mut impl FnMut(&str)) {
     loop {
         if let Some(pos) = memchr3(b'&', b'<', b'>', content.as_bytes()) {
-            let Some((chunk_before, (matched, rest))) = content
-                .split_at_checked(pos)
-                .and_then(|(before, rest)| Some((before, rest.split_at_checked(1)?)))
-            else {
+            let Some((chunk_before, rest)) = content.split_at_checked(pos) else {
                 return;
             };
+            let Some((matched, rest)) = rest.split_at_checked(1) else {
+                return;
+            };
+
             content = rest;
             let matched = matched.as_bytes()[0];
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,7 +87,9 @@ pub mod send {
 pub mod errors {
     pub use super::memory::MemoryLimitExceededError;
     pub use super::parser::ParsingAmbiguityError;
-    pub use super::rewritable_units::{AttributeNameError, CommentTextError, TagNameError};
+    pub use super::rewritable_units::{
+        AttributeNameError, CommentTextError, TagNameError, Utf8Error,
+    };
     pub use super::rewriter::RewritingError;
     pub use super::selectors_vm::SelectorError;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,7 +96,7 @@ pub mod errors {
 pub mod html_content {
     pub use super::rewritable_units::{
         Attribute, Comment, ContentType, Doctype, DocumentEnd, Element, EndTag, StartTag,
-        TextChunk, UserData,
+        StreamingHandler, StreamingHandlerSink, TextChunk, UserData,
     };
 
     pub use super::html::TextType;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,7 +206,7 @@ cfg_if! {
         };
 
         pub use self::rewritable_units::{
-            EndTag, Serialize, StartTag, Token, TokenCaptureFlags, Mutations
+            EndTag, Serialize, StartTag, Token, TokenCaptureFlags,
         };
 
         pub use self::memory::SharedMemoryLimiter;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@
 //! [Cloudflare Workers]: https://www.cloudflare.com/en-gb/products/cloudflare-workers/
 //! [`HtmlRewriter`]: struct.HtmlRewriter.html
 //! [`rewrite_str`]: fn.rewrite_str.html
+#![forbid(unsafe_code)]
 #![allow(clippy::default_trait_access)]
 #![allow(clippy::module_name_repetitions)]
 #![allow(clippy::redundant_pub_crate)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -156,6 +156,7 @@ pub mod test_utils {
 
     impl Output {
         #[must_use]
+        #[inline]
         pub fn new(encoding: &'static Encoding) -> Self {
             Self {
                 bytes: Vec::default(),
@@ -164,6 +165,8 @@ pub mod test_utils {
             }
         }
 
+        #[inline]
+        #[track_caller]
         pub fn push(&mut self, chunk: &[u8]) {
             if chunk.is_empty() {
                 self.finalizing_chunk_received = true;
@@ -179,6 +182,8 @@ pub mod test_utils {
     }
 
     impl From<Output> for String {
+        #[inline]
+        #[track_caller]
         fn from(output: Output) -> Self {
             assert!(
                 output.finalizing_chunk_received,
--- a/src/rewritable_units/mutations.rs
+++ b/src/rewritable_units/mutations.rs
@@ -1,7 +1,11 @@
 use crate::base::Bytes;
 use encoding_rs::Encoding;
+use std::error::Error as StdError;
+
+type BoxResult = Result<(), Box<dyn StdError + Send + Sync>>;
 
 /// The type of inserted content.
+#[derive(Copy, Clone)]
 pub enum ContentType {
     /// HTML content type. The rewriter will insert the content as is.
     Html,
@@ -17,7 +21,7 @@ pub(super) fn content_to_bytes(
     content: &str,
     content_type: ContentType,
     encoding: &'static Encoding,
-    mut output_handler: &mut dyn FnMut(&[u8]),
+    output_handler: &mut dyn FnMut(&[u8]),
 ) {
     let bytes = Bytes::from_str(content, encoding);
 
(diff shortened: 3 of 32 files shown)


Cargo.toml
```diff
@@ -20,6 +20,7 @@ bench = false
 
 [features]
 debug_trace = []
+# Unstable: for internal use only
```
inikulin, line 23: Why is it unstable?

src/rewritable_units/mutations.rs
```diff
@@ -27,68 +32,174 @@ pub(super) fn content_to_bytes(
...
+}
+
+impl<'tmp> StreamingHandlerSink<'tmp> {
+    /// Writes the given UTF-8 string to the output, converting the encoding and [escaping](ContentType) if necessary.
+    ///
+    /// It may be called multiple times. The strings will be concatenated together.
     #[inline]
-    pub fn removed(&self) -> bool {
-        self.removed
+    pub fn write_str_chunk(&mut self, content: &str, content_type: ContentType) {
```
inikulin, line 156: For workers use cases they want to write byte slices which makes the whole thing more complex. So, we would need streaming encoding of the written chunks

inikulin (changes requested):

src/rewritable_units/mutations.rs
kornelski: The existing API only accepts UTF-8, and Workers use that API, so I assume Workers have UTF-8 to send, and it's not a case of any-encoding to any-encoding conversion.

UTF-8 data split into arbitrary chunks just needs finding a char boundary at the end and buffering up to 3 bytes, so this can be added without major changes to the API.

IgorMinar: correct. we only need utf-8 support as that's what is used by workers and we are merging streams produced by workers/pages.

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,301 @@
...
+        match content_type {
+            ContentType::Html => self.write_html_inner(content),
+            ContentType::Text => self.write_body_text(content),
+        };
+        Ok(())
+    }
+
+    fn make_complete_utf8<'c>(&mut self, mut content: &'c [u8]) -> Option<&'c str> {
+        // Finish previous incomplete sequence if possible
+        while let Some((&first, rest)) = content.split_first() {
```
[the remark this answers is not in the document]
kornelski: Technically yes, but it's complicating things a lot for no benefit. It goes through [a big stateful decoding function for multiple encodings](https://github.com/hsivonen/encoding_rs/blob/f978f7f71311670aa068531004910eae76f789fb/src/macros.rs#L1283-L1502) which probably won't be optimized out. It needs a destination buffer to write to, which for lolhtml means extra copies.

The logic for skipping incomplete utf-8 sequence is trivial, and it doesn't need thousands of lines of extra code and lookup tables that handle a lot more complex encodings.

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,180 @@
...
+        debug_assert!(encoding != UTF_8);
+        debug_assert!(encoding.is_ascii_compatible());
+        Self {
+            encoder: encoding.new_encoder(),
+            buffer: Buffer::Stack([0; 63]),
+        }
+    }
+
+    #[inline(never)]
+    fn encode(&mut self, mut content: &str, output_handler: &mut dyn FnMut(&[u8])) {
```
inikulin, line 99: I'm confused. This just encodes a Rust string in non-UTF8 byte sequence (if streamed content is non-UTF8) and just writes it to the output sink. Is that right? If so, why not just use the previous impl: https://github.com/cloudflare/lol-html/blob/master/src/base/bytes.rs#L22?

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,362 @@
...
+            t.write_utf8_chunk(chunk, ContentType::Html).expect(&msg);
+        }
+        drop(t);
+        assert_eq!(String::from_utf8_lossy(&out), text, "{len}");
+    }
+}
+
+#[test]
+fn invalid_utf8_fragments() {
+    #[rustfmt::skip]
```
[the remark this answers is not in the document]
kornelski: because rustfmt unwraps it to one item per line, which makes it unhelpfully long.

src/rewritable_units/text_encoder.rs
kornelski: 1. `Encoding::encode()` contains UB (makes a slice from uninitialized bytes)
2. No-copy fast path is only for UTF-8 to UTF-8 and pure ASCII.
3. If there is any non-ASCII byte, it heap allocates a buffer for the entire string, and uses maximally pessimistic estimate for the buffer size, which is a 3x multiplier for CJK encodings.

My implementation doesn't have UB. It doesn't allocate unless necessary, and can reuse the buffer across multiple chunks.

src/rewritable_units/text_encoder.rs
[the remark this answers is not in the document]
inikulin: Let's rustfmt it, but just move it into a separate module file.

src/rewritable_units/text_encoder.rs
inikulin: Sounds good. Could you add a code comment explaining all that for the future generations. I imagine it will become lost knowledge otherwise

Jarred-Sumner:
nice

src/html/mod.rs
```diff
@@ -9,3 +12,63 @@ pub use self::local_name::{LocalName, LocalNameHash};
 pub use self::namespace::Namespace;
 pub use self::tag::Tag;
 pub use self::text_type::TextType;
+
+/// Convert text to HTML
+#[inline]
+pub(crate) fn escape_body_text(mut content: &str, output_handler: &mut impl FnMut(&str)) {
+    loop {
+        if let Some(pos) = memchr3(b'&', b'<', b'>', content.as_bytes()) {
```
inikulin, line 20: Nit: let's use `match` here instead of `if`

src/html/mod.rs
```diff
@@ -9,3 +12,63 @@ pub use self::local_name::{LocalName, LocalNameHash};
...
+            let Some((chunk_before, (matched, rest))) = content
+                .split_at_checked(pos)
+                .and_then(|(before, rest)| Some((before, rest.split_at_checked(1)?)))
```
inikulin, line 23: Fluent syntax is quite hard to read here. Let's use classic control flow here instead

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    #[inline]
+    pub(crate) fn output_handler(&mut self) -> &mut dyn FnMut(&[u8]) {
+        &mut self.output_handler
+    }
+}
+
+enum Buffer {
+    Heap(Vec<u8>),
+    Stack([u8; 63]), // leave a byte for the tag
```
inikulin, line 79: What's "tag" here?

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+            let ascii_len = Encoding::ascii_valid_up_to(content.as_bytes());
+            if let Some((ascii, remainder)) = content.split_at_checked(ascii_len) {
+                if !ascii.is_empty() {
+                    (output_handler)(ascii.as_bytes());
+                }
+                if remainder.is_empty() {
+                    return;
+                }
+                content = remainder;
+            }
```
inikulin, line 114: Let's refactor this part in a separate function with a meningful name that reflects that it's a fast path fro leading ASCII bytes

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+
+            let buffer = match &mut self.buffer {
+                Buffer::Heap(buf) => buf.as_mut_slice(),
+                // Long non-ASCII content could take lots of roundtrips through the encoder
+                buf if content.len() >= 1 << 20 => {
```
inikulin, line 119: Let's have a constant with a meaningful name for this magic number

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+                    *buf = Buffer::Heap(vec![0; 4096]);
+                    match buf {
+                        Buffer::Heap(buf) => buf.as_mut(),
+                        _ => unreachable!(),
+                    }
+                }
+                Buffer::Stack(buf) => buf.as_mut_slice(),
```
inikulin, line 126: Let's refactor this match into `Buffer`'s method

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+            };
+
+            let (result, read, written, _) = self.encoder.encode_from_utf8(content, buffer, false);
+            if written > 0 && written <= buffer.len() {
+                (output_handler)(&buffer[..written]);
+            }
+            if read >= content.len() {
+                return;
+            }
```
inikulin, line 135: Style nit: blank lines between statements and variable declarations

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,183 @@
...
+            content = &content[read..];
+            match result {
+                CoderResult::InputEmpty => {
+                    debug_assert!(content.is_empty());
+                    return;
+                }
+                CoderResult::OutputFull => {
+                    match &mut self.buffer {
+                        Buffer::Heap(buf) if buf.len() >= 1024 => {
```
inikulin, line 144: Let's have a constant with a meaningful name for this magic number

src/html/mod.rs
kornelski: I think this is clearer, because it checks for presence (a binary choice), and not among multiple possibilities. Using `match` for an `if/else` decision seems odd.

kornelski:
## Known issues

### `Send`

Currently the handlers use `Box<dyn StreamingHandler>` which is `trait StreamingHandler: Send`, so all streaming handlers have to be `Send` always. This is inconvenient for users of the C API, because it adds a thread-safety requirement, which the single-threaded lolhtml implementation doesn't even take advantage of. 

I can't simply remove `Send` bound from the handlers, because that makes all handlers non-`Send`, since they're all stored together in `Mutations`.

I've tried parametrizing it over `Send`, but we do that using `H: HandlerTypes` associated types, and since handlers are stored inside `Mutations`, it needs to be changed to `Mutations<S>`. This makes the type argument viral, and requires lots of tiny changes all over the place. I've used `From<F: Fn()> for Box<dyn StreamingHandler>` to support closures nicely, but when the type changes to abstract `H::StreamingHandler` all these `From`s would have to be explicit bounds on the trait `HandlerTypes` (yuck), or bounds on the methods that make type inference even more fragile. It's solvable, but needs work, so I'm leaving this for later. Single-threaded C API consumer won't cause problems in practice, since we're not going to add internal threading to lol-html anytime soon.

### Fuzzing of handlers

The fuzzers are tricky to run (may have bitrotten?). I only got AFL running, and only in a Linux VM.

The current fuzzer focuses on the parsing side, and doesn't fuzz API for adding handlers, and barely touches mutations.
I've tried to fuzz the handlers' API, but ended up with something too big for AFL to follow, so it keeps complaining it's too slow, and can't find a stable baseline: 

https://github.com/cloudflare/lol-html/commit/a28005849569987ef4236c179f357f74b6ec0bec

This probably needs to be split into many smaller fuzz targets. 

### Perf

This commit 529921b3bb224450f517dbcb2d12f62f70785715 has 11% performance hit on text-only rewriting benchmark (more complex cases are not affected). I think this is because perf of this is very sensitive to `#[inline]` picking the right boundary to inline. Anyway, other changes have improved perf by ~8%, so net difference is smaller, but it'd be nice to look at generated code to see what gets inlined in the good case, and use `#[inline(always|never)]` to enforce that in robust way.

I've used encoding_rs with a small buffer, which is an optimization for latin-based languages, but may be a pessimization for CJK languages. The buffer size could be made dynamic based on encoding if needed.

### Encoding_rs

The library requires initalized slices, which costs some performance. I've [made 0-initializing of strings faster in the standard library](https://github.com/rust-lang/rust/pull/132705), but it'd be even better to be able to use APIs that just append to an empty `String` or `Vec`. 

Unfortunately, the [functions that take `Vec` or `String` in encoding_rs trigger undefined behavior](https://github.com/hsivonen/encoding_rs/issues/79). Slices are manipulated using offsets, so one wrong offset calculation could expose uninitialized data and make a Cloudbleed again. I've submitted a couple refactorings to `encoding_rs` towards making it able to write to `MaybeUninit` slices, but it's a big library, with *a lot* of slice manipulation, so that project needs much more work.

## JS API?

The new handlers are not exposed in the JS API. I don't know if we're still maintaining it. If we do, there's a pending PR that probably should be merged first.

src/rewritable_units/text_encoder.rs
```diff
@@ -0,0 +1,460 @@
...
+                        .ok_or(Utf8Error)?;
+                    // save the incomplete bytes from the end for the next write
+                    self.char_bytes
+                        .get_mut(..invalid.len())
+                        .ok_or(Utf8Error)?
+                        .copy_from_slice(invalid);
+                    self.char_len = invalid.len() as _;
+                    // valid_up_to promises it is valid
+                    debug_assert!(std::str::from_utf8(valid).is_ok());
+                    let valid = unsafe { std::str::from_utf8_unchecked(valid) };
```
orium, line 311: Can you call `std::str::from_utf8()` instead and `.except("...")` on it (and remove the `debug_assert!()` above). I prefer to keep `unsafe` to an absolute minimum to have the least possible surface for UB.

src/rewritable_units/mutations.rs
```diff
@@ -12,83 +16,169 @@ pub enum ContentType {
...
+    fn write_all(self: Box<Self>, sink: &mut StreamingHandlerSink<'_>) -> BoxResult;
 
-        self.replacement = replacement;
-        self.remove();
+    // Safety: due to lack of Sync, this trait must not have `&self` methods
+}
+
+/// Avoid requring `StreamingHandler` to be `Sync`.
+/// It only has a method taking exclusive ownership, so there's no sharing possible.
+unsafe impl Sync for StringChunk {}
```
orium, line 155: `StreamingHandler` can be implemented for other types, including things that allow sharing and internal mutability, for instance `Arc<RefCell<Foo>>`, so taking self by moves doesn't ensure exclusivity because `self` can be a `Arc<_>`.

Do we actually need `StringChunk` to be `Sync`?

src/rewritable_units/text_encoder.rs
kornelski: Replaced with safe non-panicking code

src/rewritable_units/mutations.rs
kornelski: `Arc<T>` isn't `Send` if `T` isn't `Sync`, and `StreamingHandler` required `Send`.

I've just removed the `Sync` impl. It wasn't important.

src/lib.rs
```diff
@@ -16,6 +16,7 @@
 //! [Cloudflare Workers]: https://www.cloudflare.com/en-gb/products/cloudflare-workers/
 //! [`HtmlRewriter`]: struct.HtmlRewriter.html
 //! [`rewrite_str`]: fn.rewrite_str.html
+#![forbid(unsafe_code)]
```
orium, line 19: nice!

orium (approved):

PR #232, opened 2024-10-31
Misc cleanups

kornelski:


src/rewriter/settings.rs
```diff
@@ -82,6 +82,8 @@ impl HandlerTypes for LocalHandlerTypes {
     }
 }
 
+/// Marker type for sendable handlers. Use aliases from the `send` module.
```
orium, line 85: Intralink to the send module: ```[`crate::send`]```

orium (approved):

PR #233, opened 2024-10-31
Encoding can only be set once

kornelski:
Per HTML5, only the first one counts. 

I've also replaced `AtomicUsize` with `AtomicPtr` to avoid lookup table.


src/base/encoding.rs
```diff
@@ -1,285 +1,280 @@
...
+            // *mut T doesn't have aliasing requirements
+            encoding: Arc::new(AtomicPtr::new(ptr::from_ref(encoding).cast_mut())),
         }
     }
 
     #[must_use]
     pub fn get(&self) -> &'static Encoding {
         let encoding = self.encoding.load(Ordering::Relaxed);
-        ALL_ENCODINGS[encoding]
+        unsafe { &*encoding }
```
inikulin, line 30: Can we not have `unsafe` blocks for that? I'm fine with a lookup table as a price for that

src/selectors_vm/attribute_matcher.rs
```diff
@@ -50,10 +50,14 @@ impl<'i> AttributeMatcher<'i> {
...
-                let attr_name = self.input.slice(a.name);
-
-                for i in 0..attr_name.len() {
-                    if attr_name[i].to_ascii_lowercase() != lowercased_name[i] {
+                let attr_name = &*self.input.slice(a.name);
+
+                for (attr, wanted) in attr_name
+                    .iter()
+                    .copied()
+                    .zip(lowercased_name.iter().copied())
```
inikulin, line 58: Why this change?

src/rewriter/mod.rs
```diff
@@ -266,22 +261,29 @@ impl<'h, O: OutputSink, H: HandlerTypes> Debug for HtmlRewriter<'h, O, H> {
 fn handler_adjust_charset_on_meta_tag<'h, H: HandlerTypes>(
     encoding: SharedEncoding,
 ) -> (Cow<'h, crate::Selector>, ElementContentHandlers<'h, H>) {
+    let mut set_once = false;
```
inikulin, line 264: Nit: let's rename it to `found` and also add a blank line after the variable declaration

src/selectors_vm/attribute_matcher.rs
kornelski: Generates much leaner code

src/base/encoding.rs
kornelski: What's wrong with that?

The pointers have a static lifetime, so it's not a risky code.

inikulin (approved):

PR #235, opened 2024-11-01
C API improvements

kornelski:


c-api/src/comment.rs
```diff
@@ -1,12 +1,12 @@
 use super::*;
 
 #[no_mangle]
-pub extern "C" fn lol_html_comment_text_get(comment: *const Comment) -> Str {
+pub unsafe extern "C" fn lol_html_comment_text_get(comment: *const Comment) -> Str {
```
inikulin, line 4: Introducing this change, let's set `#![forbid(unsafe_op_in_unsafe_fn)]` for the crate

inikulin (changes requested):

c-api/src/comment.rs
kornelski: Not ready to do that yet, as it will cause merge conflicts for the streaming handlers branch.

orium (approved):

Issue #216, opened 2024-07-31
Consider upgrading to a current cssparser


fasterthanlime:
Right now, `cssparser@0.27.2` pulls in `syn@1` through `phf_macros@0.8` — it's not great:

```
❯ cargo tree -i syn@1
syn v1.0.109
├── derive-try-from-primitive v1.0.0 (proc-macro)
│   └── lith v4.0.0 (/Users/amos/bearcove/lith/app)
└── phf_macros v0.8.0 (proc-macro)
    └── phf v0.8.0
        ├── cssparser v0.27.2
        │   ├── lol_html v1.2.1
        │   │   └── lith v4.0.0 (/Users/amos/bearcove/lith/app)
        │   └── selectors v0.22.0
        │       └── lol_html v1.2.1 (*)
        └── selectors v0.22.0 (*)
[build-dependencies]
└── cssparser v0.27.2 (*)
```

syn is an extremely costly dependency:

![CleanShot 2024-07-31 at 19 21 41@2x](https://github.com/user-attachments/assets/7034fc43-b964-4021-9538-4115f210d303)

I'd rather have only one of them — preferably, zero.

`cssparser@0.34` pulls in `phf@0.11` which depends on `syn@2`.

kornelski:
https://github.com/cloudflare/lol-html/pull/250

PR #238, opened 2024-11-06
Update `cssparser` to v0.28 and `selectors` to v0.23

antonok-edm:
This makes incremental progress towards https://github.com/cloudflare/lol-html/issues/216, although there are still some versions to go before `syn` v1 can be removed from the dependency tree.


kornelski:
I'll merge it after the next release, to minimize amount of changes per release.

kornelski (approved):

PR #241, opened 2024-11-11
Streaming handlers chunked

kornelski:
Ability to write chunks of UTF-8 text split at arbitrary positions.

It's on top of: https://github.com/cloudflare/lol-html/pull/229


npaun (approved):
LGTM from the point of view of a consumer of this API

PR #247, opened 2024-12-04
c-api: use cargo-c

selfisekai:


selfisekai:
The generated header differs a lot from the removed one for some reason

kornelski:
I've had the same issue with cargo-c. I think by default it wants to generate a new header using cbindgen. Our hand-written header is much better.

You can use this as a template to switch to hand-written header:

https://github.com/ImageOptim/libimagequant/blob/fea449e31038a1aeddf85b768d90426bb3a0411c/imagequant-sys/Cargo.toml#L42C1-L45

kornelski (approved):

kornelski:
Thank you!

PR #248, opened 2024-12-04
Upgrade dependencies

kornelski:


src/selectors_vm/ast.rs
```diff
@@ -49,15 +49,15 @@ impl NthChild {
 pub(crate) enum OnTagNameExpr {
     ExplicitAny,
     Unmatchable,
-    LocalName(String),
```
orium, line 52: Why are we replacing `Strings` with `Box<str>`?

src/rewriter/mod.rs
```diff
@@ -354,13 +354,15 @@ mod tests {
...
-    use static_assertions::assert_impl_all;
     use std::convert::TryInto;
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
 
     // Assert that HtmlRewriter with `SendHandlerTypes` is `Send`.
-    assert_impl_all!(crate::send::HtmlRewriter<'_, Box<dyn FnMut(&[u8]) + Send>>: Send);
+    const _: () = {
+        const fn is_send<T: Send>() {}
+        is_send::<crate::send::HtmlRewriter<'_, Box<dyn FnMut(&[u8]) + Send>>>()
```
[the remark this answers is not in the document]
kornelski: Because we can avoid pulling in an entire crate to get exactly this code. Static assertions used to be very tricky in Rust and needed odd hacks, but since `const` eval has improved, the `static_assertions` crate is largely obsolete.

src/selectors_vm/ast.rs
kornelski: I've noticed that these strings aren't grown anywhere, and are always created from `&str`, so we're carrying a redundant capacity for no reason.

Newer versions of `selectors` require storing a precomputed hash, so this makes room for adding it without regressing struct sizes.

orium: Sounds good.

src/rewriter/mod.rs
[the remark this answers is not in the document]
orium: This is a dev dependency so it's not adding any extra work for crates depending on lol-html. If I saw this code I would replace it with `static_assertions` and say something like "Simplified the const check by using `static_assertions`.".

orium (approved):

PR #253, opened 2024-12-24
Make LocalNameHash smaller

kornelski:
It's a bit ironic that `LocalNameHash` tries so hard to compress 12-byte names into 64 bits, but `Option<u64>` makes it 16 bytes large.

I've changed `Option` to `-1` to keep it 64-bit.


kornelski:
Since the hash is reversible, I've also added a useful `Debug` impl for it.

orium (approved):

scotchmist (approved):

PR #263, opened 2025-03-02
Add element.onEndTag js binding

taina0407:
This PR is to support onEndTag for JS API.
This is the first time I'm working on a Rust project so feel free to suggest any changes

### DIFF
--- a/js-api/Cargo.lock
+++ b/js-api/Cargo.lock
@@ -192,7 +192,7 @@ checksum = "a7a70ba024b9dc04c27ea2f0c0548feb474ec5c54bba33a7f72f873a39d07b24"
 
 [[package]]
 name = "lol-html"
-version = "2.1.1"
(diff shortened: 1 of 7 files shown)


kornelski (approved):

kornelski:
There's only one EndTag type, so the macro for defining mutation methods seems unnecessary?

taina0407:
> There's only one EndTag type, so the macro for defining mutation methods seems unnecessary?

You are right. The problem here is EndTag seems missing `removed` property, so I changed the PR to adding this so we can re-use existing `impl_mutations` macro instead.

kornelski:
Looks good. Small nit: the comment for removed is a copypaste error

taina0407:
@kornelski thankyou, amend the commit to change "comment" to "end tag" :)

PR #266, opened 2025-04-23
chore: update lockfile

chenrui333:
seeing outdated cargo.lock when building lol-html 2.3.0 release

relates to https://github.com/Homebrew/homebrew-core/pull/220967


kornelski (approved):

kornelski:
Does Brew need the lock file? We could delete it instead.

chenrui333:
yeah, we do use the lockfile (it helps with the reproducible build)

PR #270, opened 2025-06-13
Borrowed Bytes

kornelski:
Owned bytes are not required everywhere. This makes structs smaller, and everything slightly faster.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "lol_html"
-version = "2.4.0"
+version = "2.5.0"
 authors = ["Ivan Nikulin <inikulin@cloudflare.com, ifaaan@gmail.com>"]
 license = "BSD-3-Clause"
 description = "Streaming HTML rewriter/parser with CSS selector-based API"


src/base/bytes.rs
```diff
@@ -10,27 +10,27 @@ use std::str;
...
 pub struct HasReplacementsError;
 
+/// A thin wrapper around byte slice with handy APIs attached
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+#[repr(transparent)]
+pub(crate) struct Bytes<'b>(&'b [u8]);
+
 /// A thin wrapper around either byte slice or owned bytes with some handy APIs attached
 #[derive(Clone, PartialEq, Eq, Hash)]
 #[allow(unnameable_types)] // accidentally exposed via `tag.set_name()`
```
orium, line 20: instead of the comment I think you can do 

```rust
#[allow(unnameable_types, reason = "accidentally exposed via `tag.set_name()`")]
```

src/parser/mod.rs
```diff
@@ -62,6 +62,7 @@ impl<S: ParserOutputSink> Parser<S> {
         }
     }
 
+    #[inline(never)]
```
orium, line 65: Can you add a comment explaining why this should never be inlined?

src/rewritable_units/tokens/end_tag.rs
```diff
@@ -49,20 +54,20 @@ impl<'i> EndTag<'i> {
     #[inline]
     #[doc(hidden)]
     #[deprecated(note = "use set_name_str")]
-    pub fn set_name(&mut self, name: Bytes<'static>) {
+    pub fn set_name(&mut self, name: BytesCow<'static>) {
```
orium, line 57: Was this meant to be `pub(crate)`?

kornelski: This is kept for semver-compat. It should have taken `&str` instead of a type that cannot be named and exposes internals.

PR #286, opened 2025-08-26
Fix some oversights in #285

hexchain:
Sorry for the trouble!


hexchain:
@briancarbone Indeed, thanks! This is also fixed.

PR #307, opened 2026-03-11
Reallocs

kornelski:
The compiler has been copying every single string from the `Ast`, but it doesn't have to.

The BytesCow API was unclear about borrowing vs copying, so I've fixed that. Since that was affecting an accidentally-public API, it depends on a cleanup: https://github.com/cloudflare/lol-html/pull/311


kornelski:
the semver-checks failure is expected and intentional. I've fixed `set_name`, so `set_name_str` previous replacement can be hidden now. https://github.com/cloudflare/lol-html/pull/311

orium (approved):

PR #313, opened 2026-04-23
Add name_source_location() and value_source_location() to Attribute

gmalette:
Exposes document-absolute byte ranges for attribute names and values via the existing SourceLocation type.

We need this to perform byte-level splice rewrites on attribute values without re-parsing the start tag... I know we're holding it wrong but this is the fastest and most correct way to achieve our goals :)

### DIFF
--- a/src/parser/lexer/lexeme/mod.rs
+++ b/src/parser/lexer/lexeme/mod.rs
@@ -39,6 +39,11 @@ impl<'i, T> Lexeme<'i, T> {
     }
 
     #[inline]
+    pub const fn input_byte_offset(&self) -> usize {
+        self.previously_consumed_byte_count
+    }
+
+    #[inline]
     pub const fn token_outline(&self) -> &T {
         &self.token_outline
     }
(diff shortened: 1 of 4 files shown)


kornelski:
Thanks for the PR.

The change isn't too bad, but it makes me wonder why aren't you using the _rewriter_ to do rewriting?

kornelski (approved):

gmalette:
> but it makes me wonder why aren't you using the rewriter to do rewriting?

Yes this is a very good question and I should've led with that. 

The reason is: we rewrite html bodies not written by us. We used to do this with regexps and it was very imprecise and so often we targeted nodes that were unrelated to the *reason* for the rewrite. For example, if your body contained `type="importmap"`, we injected the polyfill script before the very first `<script>`, even if it was unrelated.

Changing the payload we output is like turning a cargo ship around, and while we want to move to full `rewriter`, right now we use it also as a scanner for those cases. We register node positions and perform the string manipulation when it's all computed.

It's not great and I know it's not how you want us to use it, but it's also 2-7x faster and approximately infinity times more precise.

gmalette:
@kornelski would you cut a new release please? 🙏

PR #320, opened 2026-05-20
Add opt-in graceful bail-out for memory and content handler errors

orium:
Today a `MemoryLimitExceeded` or `ContentHandlerError` aborts the rewriter mid-stream and leaves
the downstream response truncated.

This PR adds two opt-in flags:

1. `MemorySettings::graceful_bail_out_on_memory_limit_exceeded`
2. `Settings::graceful_bail_out_on_content_handler_error`

When set, before returning the error the rewriter flushes every input byte it has received but
not yet emitted to the sink, as-is. The caller can continue the response by writing subsequent
bytes directly to its downstream sink. Both flags default to `false`: existing behavior is
unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## v3.0.0
 
 - Added `MemorySettings::graceful_bail_out_on_memory_limit_exceeded`: when set, the rewriter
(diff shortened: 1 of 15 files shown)


kornelski (approved):
This will need non_exhaustive and semver-major bump.

PR #329, opened 2026-07-25
Fix selector matching being skipped for consecutive integration-point siblings

robobun:
## The bug

When two SVG/MathML integration-point elements are adjacent siblings, the second one never reaches selector matching. It is emitted to the output verbatim, but no element handler runs for it, including the universal selector:

```rust
// handler for "foreignobject" (or "*")
rewrite_str(r#"<svg><title>x</title><foreignObject></foreignObject></svg>"#, settings)
// handler fires for svg and title, never for foreignObject
```

Remove the `<title>` sibling and the same `foreignObject` matches fine. Intervening text does not re-arm matching; only an intervening non-integration-point element does. This makes selector-based sanitizers built on lol-html bypassable by prefixing the payload element with another integration-point sibling.

## The cause

The tag scanner emits a tag hint for most start tags; `Dispatcher::apply_capture_flags_from_hint_and_get_next_parser_directive` sets `got_flags_from_hint` so the lexer's `handle_tag` does not re-run selector matching for a tag the hint already matched. Integration-point start tags take a different path: the tree-builder simulator answers `RequestLexeme` (it needs `self_closing` to decide namespace entry) and the scanner switches to the lexer without emitting a hint.

If the previous tag's hint left the parser in scan mode (nothing to capture), `got_flags_from_hint` stayed `true`. The next `RequestLexeme` tag then reached `handle_tag` with the stale flag, so `adjust_capture_flags_for_tag_lexeme` was skipped and `handle_start_tag` never ran for it.

## The fix

The flag is only meaningful when the hint results in an immediate switch to the lexer for the same tag, so set it to `matches!(directive, ParserDirective::Lex)` instead of unconditionally `true`.

Two regression tests added next to the existing foreign-content tests in `rewritable_units/element.rs` (SVG `desc/title/foreignObject` and MathML `mi/mo/mn/annotation-xml` sibling chains); both fail on main and pass with the fix. `cargo test` and `cargo test --features=_integration_test` are green.

Found while debugging Bun's HTMLRewriter (oven-sh/bun#35673, ships this change as a vendored patch).

### DIFF
--- a/src/rewritable_units/element.rs
+++ b/src/rewritable_units/element.rs
@@ -978,6 +978,38 @@ mod tests {
     }
 
     #[test]
+    fn consecutive_svg_integration_points() {
+        // A tag hint that keeps the parser in scan mode must not leave
+        // `got_flags_from_hint` set: the next integration-point tag enters the
+        // lexer via tree-builder feedback without a hint of its own, and the
+        // stale flag would make `handle_tag` skip selector matching for it.
+        let out = rewrite_element(
+            b"<svg><desc>d</desc><title>t</title><foreignObject></foreignObject></svg>",
+            UTF_8,
+            "desc,title,foreignobject",
+            |el| el.set_attribute("hit", "").unwrap(),
+        );
+        assert_eq!(
+            out,
+            r#"<svg><desc hit="">d</desc><title hit="">t</title><foreignObject hit=""></foreignObject></svg>"#
+        );
+    }
+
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (changes requested):
@robobun add a test and make sure you match the style of other commits and any lints in this repo

orium (approved):

